PR #1, opened 2018-07-21
first iteration of ccm 

ashleyschuett:
You can make an image test out the ccm with `make build-ccm` 

The needed deployments are under `releases/*`, just need to change the data for the secret to be your api key and project id. I wasn't sure how the versioning was going to work for this so I just have it as 0.0.0.


packet/devices.go
```diff
@@ -0,0 +1,191 @@
...
+
+// InstanceID returns the cloud provider ID of the node with the specified NodeName.
+// Note that if the instance does not exist or is no longer running, we must return ("", cloudprovider.InstanceNotFound)
+func (i *instances) InstanceID(_ context.Context, nodeName types.NodeName) (string, error) {
+	device, err := deviceByName(i.client, i.project, nodeName)
+	if err != nil {
+		return "", err
+	}
+
+	return device.ID, nil
```
ntfrnzn: The providerID has to match the expected format, perhaps by changing this here to
```
   return fmt.Sprintf("packet://%s", device.ID), nil
```

ntfrnzn:
Seems to be a good framework. I'd prefer to build the binary in a golang builder container and then copy it, rather than ship the golang build environment, but that's minor.

PR #19, opened 2019-10-28
Update Makefile and switch to GH Actions

deitch:
Align the Makefile with what we do in CSI, and update from Travis to GH Actions


deitch:
Nah, the errors unfortunately were not trivial. Bogged me down for a good few hours. I will get CI clean, and then update it here. Close....

PR #24, opened 2019-11-12
Patch for GCE flag issue

### LINKED ISSUE #23: ccm crashes due to gce flag
[This bug](https://github.com/kubernetes/kubernetes/pull/76627) is manifesting itself. Either fix it manually in `main.go` as [digital ocean did](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/209/files#diff-a29863390f1b096d59524d3e76ad507dR38-R40) or upgrade to k8s 1.15.0 or higher, which includes that fix, as [DO did](https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/238)

Ideally we will get 1.15 working, and this will go away.

deitch:
Fixes #23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ ifndef PKG_LIST
 	$(eval PKG_LIST := $(shell $(BUILD_CMD) go list ./... | grep -v vendor))
 endif
 
-.PHONY: fmt-check lint test vet golint tag version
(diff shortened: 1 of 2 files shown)


deitch:
Also added `Makefile` target to format files

deitch:
This is no longer needed, as #26 covers it. Closing.

PR #32, opened 2019-11-26
Support secrets from file to align with CSI

### LINKED ISSUE #22: Unify secrets with how csi does it
[packet-csi](https://github.com/packethost/csi-packet) supports env vars, but also config file passed via `--config=/...`. This should be unified between the two.

deitch:
Fixes #22 

**Do not merge until [WIP] is removed.**

This is a first run. There still are some changes to be made, but they depend on #31 being merged first. I will remove `[WIP]` and comment that it is ready, when you can merge it in.

This PR aligns secrets with how [csi-packet](https://github.com/packethost/csi-packet) does it, allowing a single secret to serve both. It will be our standard going forward.

When this is done, the api key and project ID can be in one of either of the following two places (aligned with CSI):

* env vars, as per the current implementation
* a secrets file in json structure

Eventually, we will allow it to dynamically monitor the secrets file and update if it changes.

What is left to complete this:

- [x] Merge #31 and rebase off of it
- [x] Update `template/secret.yaml` to reflect the new design
- [x] Update documentation to describe the new design
- [x] Test again on a live cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,11 @@ metadata:
   name: packet-cloud-config
   namespace: kube-system
 stringData:
-  apiKey: "abc123abc123abc123"
-  projectID: "abc123abc123abc123"
+  cloud-sa.json: |
+    {
+    "apiKey": "abc123abc123abc123"
+    "projectID": "abc123abc123abc123"
+    }  
 ```
 
 Then apply the file via `kubectl`, e.g.:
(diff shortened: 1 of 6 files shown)


deitch:
FYI @jmarhee this _will_ cause some minor changes to the Helm charts. Don't worry about it until we are in.

PR #34, opened 2020-02-02
Implement metallb load balancer, BGP, node and services watchers

deitch:
**Do not merge this.**

This PR currently only contains doc changes, not the actual implementation. I will remove the `[WIP]` and comment when it is ready to be merged.

I am looking for feedback on the design of loadbalancer support in CCM. The goal is that customers who deploy the packet CCM get load balancers for `Service` of `type=LoadBalancer` that "just works".

I have tested this process manually and it looks clean, but I want input on the design before going ahead to implementation.


README.md
```diff
@@ -99,3 +99,52 @@ optional additional logging levels via the `--v=<level>` flag. In general:
...
+
+The process is as follows goes through each `Service` and the tagged IP address reservations.
+
+1. If there are `Service` with IP pending, and there are available tagged Reservations, add them to the `ConfigMap`
+1. If there remain `Service` with IP pending, and there are no available tagged Reservation or Request, create the request; if it responds immediately, add them
+1. At this point, `count(Service) <= count(Reservation+Request)`
+1. If `count(Requests+Reservations) > count(Service)`, delete Requests until the count is the same, or there are no more Requests
+1. If `count(Reservations) > count(Service)`:
+   1. find unused IPs in the `ConfigMap` by comparing with `Service`s and remove them from the `ConfigMap`
+   1. delete Reservations for those IPs
```
pereztr5: Do we think that auto cleaning up the unused IPs is a good thing? I can see situations where you want to delete the service but keep the IP since that is something that you cannot usually change quickly like in DNS.

pereztr5:
Besides my comment around the auto cleaning up of unused Elastic IPs things look good for me.

README.md
deitch: I think so, yes, if we can tag them appropriately. DNS may be quick to change, but it actually is a lot slower to propagate (especially if you have a longer TTL, or clients that don't respect it). 

The other consideration is cost. Customers pay for those IPs, and so if they are created automatically when a `Service` is created, they should be deleted when the `Service` is deleted.

I definitely am open to adding an option later to disable it, though. :-)

deitch:
Update: this now contains most of the implementation for load balancers via metallb. It currently is awaiting a few things:

* lots more tests and testing
* ability to get IPs by device, so that we can complete [this](https://github.com/packethost/packet-ccm/pull/34/files#diff-18806f1370c3b4658d7356de518e80d6R246-R248). It depends on [this packngo PR](https://github.com/packethost/packngo/pull/168) which should be in later this week
* two not-yet-existing features for IP reservation requests that are not automatically approved:
   * ability to track them via API+packngo
   * ability to retain tags from request through to reservation

The last one **is a blocker**. We need it so we can know that service X had a reservation request, and thus we should not submit a new one, and when it completes, we can affiliate it with the appropriate service. Otherwise, we will end up submitting a new request for every such service every 60 seconds on loop, which is "a bad thing".

deitch:
Update: this PR has changed quite a bit. Here is what it does now:

* sets up metallb, can be disabled by config. For now it is **hard-coded disabled** until the IP API issues listed [here](https://github.com/packethost/packet-ccm/pull/34#issuecomment-588121896) are resolved
* enables BGP on the project and on each host
* creates a node watcher and a services watcher, and enables any component to be called with them

The last element lays the groundwork for #39 , which will be simpler after this is done.

This **is** ready to be merged in.

PR #36, opened 2020-02-24
README.md: add missing comma to secret example

mpolednik:
If secret was copied as-is, without comma, Packet CCM would be stuck
in a crash loop:

provider config error: failed to process json of configuration file at
path /etc/cloud-sa/cloud-sa.json: invalid character '"' after object
key:value pair

Signed-off-by: Martin Polednik <m.polednik@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ metadata:
 stringData:
   cloud-sa.json: |
     {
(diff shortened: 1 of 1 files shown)


deitch:
Nice catch @mpolednik; thank you!

PR #53, opened 2020-07-07
enable metallb

deitch:
This largish PR does the following:

* enables load balancers using metallb by default on Packet CCM. It can be disabled by option, as listed in the README [here](https://github.com/packethost/packet-ccm#load-balancers)
* ensures it all syncs correctly
* adds a `make fmt` command that formats the files
* updates the README doc for running locally (usually during development) with the manifests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ ifndef PKG_LIST
 	$(eval PKG_LIST := $(shell $(BUILD_CMD) go list ./... | grep -v vendor))
 endif
 
-.PHONY: fmt-check lint test vet golint tag version
+.PHONY: fmt fmt-check lint test vet golint tag version
 
 $(DIST_DIR):
 	mkdir -p $@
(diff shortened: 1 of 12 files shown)


deitch:
Some important notes about the load balancer

* By default, a metallb speaker runs on all nodes, including control plane nodes. We may want to provide the user with an option to disable it. If so, future PR
* I haven't looked into using it also for a control plane load balancer, or how it would conflict with one

gianarb:
ClusterAPi uses the spec released under `./deploy/releases` to deploy packet-ccm, will it be updated as part of the release process?

deitch:
> ClusterAPi uses the spec released under ./deploy/releases to deploy packet-ccm, will it be updated as part of the release process?

Yes. Once this is merged in, we should cut a new release (I think v1.1.0 per semver) and then we can consume it.

displague:
https://github.com/packethost/packet-ccm/pull/56 adds a command `make prerelease` to prepare the releases.

PR #56, opened 2020-07-07
Improve deployment release process

### LINKED ISSUE #54: Report UserAgent in API calls

### LINKED ISSUE #55: Document or script release specific changes
Missing release artifacts are easy when the process is not documented, scripted, or automated. The current release of packet-ccm is v1.0.1, however the latest [deployment recommended is v1.0.0](https://github.com/packethost/packet-ccm/blob/v1.0.1/README.md#deploy-project-and-api).  The [`deploy/release` path](https://github.com/packethost/packet-ccm/tree/v1.0.1/deploy/releases) was not updated in the latest release.

Update the documentation or scripts to make release deployment a better process for the engineer performing the steps.

displague:
_Rethinking to take https://github.com/packethost/packet-ccm/pull/56#issuecomment-656157033 into consideration_

This PR Improves CCM deployment releases.

* Adds `make prerelease RELEASE_VERSION={RELEASE_VERSIOM}`
  * revises URLs in README.md
  * creates new `deploy/release/{RELEASE_VERSION}` files (from a template)
* Updates `BUILD.md` on the use of `make prerelease`
* Markdown linting fixes were applied to BUILD.md for readability.

Closes #55

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,53 +1,60 @@
 # CCM for Packet Build and Design
+
 The Cloud Controller Manager (CCM) Packet plugin enables a Kubernetes cluster to interface directly with
 Packet cloud services.
 
 ## Deploy
+
 Read how to deploy the Kubernetes CCM for Packet in the [README.md](./README.md)!
 
 ## Building
+
 To build the binary, run:
 
-```
+```console
 make build
 ```
 
 It will deposit the binary for your local architecture as `dist/bin/packet-cloud-controller-manager--$(OS)-$(ARCH)`
 
-By default `make build` builds the binary using your locally installed go toolchain.
+By default `make build` builds the binary using your locally installed go tool-chain.
 To build it using a docker container, do:
 
-```
+```console
 make build DOCKERBUILD=true
 ```
 
 By default, it will build for your local operating system and CPU architecture. You can build for alternate architectures or operating systems via the `OS` and `ARCH` parameters:
 
-```
+```console
 make build OS=darwin
 make build OS=linux ARCH=arm64
 ```
 
 ## Docker Image
+
 To build a docker image, run:
 
-```
+```console
 make image
 ```
 
 The image will be tagged with `:latest`
 
 ## CI/CD/Release pipeline
+
 The CI/CD/Release pipeline is run via the following steps:
 
 * `make ci`: builds the binary, runs all tests, builds the OCI image
 * `make cd`: takes the image from the prior stage, tags it with the name of the branch and git hash from the commit, and pushes to the docker registry
+* `make prerelease`: updates README.md and creates deployment manifests
 * `make release`: takes the image from the `ci` stage, tags it with the git tag, and pushes to the docker registry
 
 The assumptions about workflow are as follows:
 
 * `make ci` can be run anywhere. The built binaries and OCI image will be named and tagged as per `make build` and `make image` above.
 * `make cd` should be run only on a merge into `master`. It generally will be run only in a CI system, e.g. [drone](https://drone.io) or [github actions](https://github.com/features/actions). It requires passing both `CONFIRM=true` to tell it that it is ok to push, and `BRANCH_NAME=${BRANCH_NAME}` to tell it what tag should be used in addition to the git hash. For example, to push out the current commit as master: `make cd CONFIRM=true BRANCH_NAME=master`
+* `make prerelease` should be run locally. It requires passing `RELEASE_TAG=${RELEASE_TAG}` to tell it what tag to use when updating `README.md` and creating `deploy/release` files.
 * `make release` should be run only on applying a tag to `master`, although it can run elsewhere. It generally will be run only in a CI system. It requires passing both `CONFIRM=true` to tell it that it is ok to push, and `RELEASE_TAG=${RELEASE_TAG}` to tell it what tag this release should be. For example, to push out a tagged version `v1.2.3` on the current commit: `make release CONFIRM=true RELEASE_TAG=v1.2.3`.
 
 For both `make cd` and `make release`, if you wish to push out a _different_ commit, then check that one out first.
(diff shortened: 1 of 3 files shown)


deitch:
Thanks for this @displague . It is quite appreciated when someone jumps in to help. 😀

I have a few questions around this:

* You wrote "When a particular commit is ready to cut a release, **on master**, run `make prerelease`, stage (`git add`) the new files and commit the changes". That won't work as `master` is a protected branch. We only do tags that way because, well, GitHub doesn't support PRs for tags. I would prefer it if they did (gitlab does), but so be it. How would we do this?
* I like the template idea, but what happens when the template changes? It should be fairly consistent for now, as k8s apis are pretty stable (finally), but I can see it changing over time, especially as we add capabilities.

displague:
> I like the template idea, but what happens when the template changes?

We change the template and the next release (`make prerelease`) will be derived from that template.  The old release files will still exist in the old template (with the old limitations of which versions of K8s they are expected to work well with).

We could extend this behavior to have a template for each significant Kubernetes version.  This would allow us to offer alternatives:
* `deploy/release/v1.2.0/deployment-k8s-1.11.yaml`
* `deploy/release/v1.2.0/deployment-k8s-1.15.yaml`

If it weren't for the tricky (early) timing needed to install the CCM, I would suggest a [`kustomize.yaml` approach](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/).  I don't think many users will find their clusters in a state where they can run `kubectl apply -k` (preferring tools that have bundled the CCM).  Those tools could adopt Kustomize, some more easily than others (like [cluster-api-provider-packet](https://github.com/kubernetes-sigs/cluster-api-provider-packet/blob/ab54c495108472b3989fcb581517be10790367e7/templates/cluster-template.yaml#L28-L30)).

displague:
> That won't work as master is a protected branch. We only do tags that way because, well, GitHub doesn't support PRs for tags. I would prefer it if they did (gitlab does), but so be it. How would we do this?

I didn't realize the master branch was protected here. I can change the wording here to a 'release branch' or 'release PR'.
This branch or PR is one that is ushering in updated documentation for the new version.

I wanted to avoid making `git` commands part of the Makefile.
I imagine the process would look like this:

```bash
VER=v1.2.0
git fetch origin; git checkout origin/master
git checkout -b release-$VER
# any last-minute template changes
make prerelease RELEASE_VERSION=$VER
# any other documentation changes
git add deploy/release/$VER
git commit -a -m "$VER"
git push origin release-$VER
```

I could include this in a `hack/prerelease.sh` script if that helps. 

The merge for this PR should then be a `--ff-only` one ([Rebase and merge](https://docs.github.com/en/github/administering-a-repository/about-merge-methods-on-github#rebasing-and-merging-your-commits)) to avoid the extra merge commit (but that isn't necessary, just a matter of aesthetics).

After PR merge, the Github workflow would then be triggered to stamp the tag.

How does this sound?

/cc @gianarb

displague:
While we are discussing `make prerelease`, feel free to cherry-pick the other two commits in this PR for the upcoming release.
I can spin these out into new PRs if preferred.

deitch:
> I can spin these out into new PRs if preferred.

That would be great. Those 2 commits make sense all by themselves, so let's get those into a PR and merged immediately. Then we can deal with `make prerelease`.

deitch:
> How does this sound?

You basically are trying to avoid someone having to do 2 steps - PR for the release artifacts in `deploy/` and `git tag` - but since now we have nothing to automate it, I see anything you do as an improvement.

> I didn't realize the master branch was protected here

Yup.

> I wanted to avoid making git commands part of the Makefile.
> I imagine the process would look like this:

When would you run this? What would the user flow be?

Conversely, could we have the CI process just take care of creating the artifacts, so it doesn't even need a PR? It is handled [here](https://github.com/packethost/packet-ccm/blob/master/.github/workflows/ci.yaml#L46-L48), so could we make running the steps part of the release?

We wouldn't want to change the repo (could we even?) but perhaps the very yaml manifest could be a _release artifact_ that ends up on the releases page? And our README just points to it?

gianarb:
> but perhaps the very yaml manifest could be a release artifact that ends up on the releases page? And our README just points to it?

I like this! 

> could we have the CI process just take care of creating the artifacts, so it doesn't even need a PR? 

It will be great to know that in order to get a release you just have to push a new tag! 🤯

displague:
I'll spin up the new PRs now.

> We wouldn't want to change the repo (could we even?) but perhaps the very yaml manifest could be a release artifact that ends up on the releases page? And our README just points to it?

That's perfect.  If we do this we can remove the `deploy/release/` paths.  This would break ClusterAPI unless we pin https://github.com/kubernetes-sigs/cluster-api-provider-packet/blob/ab54c495108472b3989fcb581517be10790367e7/templates/cluster-template.yaml#L30 to a tag instead of `master` (@gianarb).  I can send that PR out as well.

deitch:
Fair price to pay!

displague:
This will take a little more tinkering. Getting assets included in the release, with our current workflow, probably involves https://github.com/marketplace/actions/create-a-release and https://github.com/marketplace/actions/upload-a-release-asset (unless https://github.com/marketplace/actions/release-drafter can do provide assets).

`make prerelease` can be modified to only produce the new chart (`make chart`?), and this can become one action of `make release`.  The artifact can then be captured by the upload-a-release-artifact Workflow action, and the create-a-release action can include the install instruction (with the specific version included).

deitch:
Where does this sit @displague ? We need to cut a new release soon, would like to get this in.

displague:
@deitch I haven't revisited this, I would recommend performing a release in the traditional way until we can flesh this one out more.

deitch:
Ok thanks.

deitch:
@displague #66 handles the automation part. I backfilled the earlier releases manually. If there are other non-release-related issues (linting and such) here, can you open a new PR for it? 

Also, can we now close this one out?

PR #58, opened 2020-07-09
Control Plane Endpoint manager and reconciliation loop

### LINKED ISSUE #54: Report UserAgent in API calls

### LINKED ISSUE #57: Elastic IP management for control plane
Initially, this issue got discussed in [CAPP #141](https://github.com/kubernetes-sigs/cluster-api-provider-packet/issues/141).

## The problem statement: No Load balancer

Packet does not provide a load balancer, it means that you have to figure out a way to route traffic between Control Planes when running Kubernetes in HA otherwise you will end up having a single point of failure (SPOF) even if you run k8s HA. Because only the Node that has the Elastic IP attached to it gets traffic.

When looking at how other cloud providers solve this issue it is clear that an externally managed load balancer will make this implementation straightforward. Your cluster gets a load balancer and all the control planes are registered there.

## But we do not have a load balancer

We evaluated a couple of different ways originally to get around this problem, I will summarize them here

1. BGP at Control Plane device level with bird or equivalent. This will may work but it will conflict with MetalLB. The load balancer implementation that we ship with Kubernetes via CCM #53 because we will end up having two BGP communications ongoing from the same node, one from bird and one from MetalLB. This can potentially confuse the router.

2. MetalLB for Control Plane. This is a nice and clean solution in theory, but in practice, it is hard to implement. You need to have a  control plane up and running in order to place a Service type=LoadBalancer (managed by MetalLB) in front of the API Servers. But you can not have a reachable API Server until you get the right ElasticIP assigned to the Load Balancer. 🔁 

3. Write a control loop based on time (every 6sec for example) that will check the status of the ElasticIP, if it does not respond it can be attached to another control plane because it means that the current one is broken.

We discussed the pros and cons in the other issue linked above.

## This is a CCM issue... maybe

We are moving the conversation here because we think this is a better home for this kind of feature. Kubernetes HA on Packet is something that does not require ClusterAPI. Kubernetes clusters can be provisioned via Terraform or other tools. Fixing this issue here will make everybody using Kubernetes on Packet and CCM capable of using this feature.

gianarb:
Fixed #57 

We decided to go with a reconciliation loop written in Go that will healthcheck the Control Plane Endpoint. If it does not respond it will identify a valid control plane and it re-assign Elastic IP.


deitch:
This looks pretty good. I see a few questions:

* it doesn't fix it yet, just returns nil, but I think you knew that; this is just the framework
* it says it fixes #54 , but this has nothing to do with 54? 
* I don't see a default for the tag, which we should have
* docs on how to configure it (README, mainly)

gianarb:
> it doesn't fix it yet, just returns nil, but I think you knew that; this is just the framework

Right :) half an hour after we agreed on the plan is not enough for me to have it running 🍡 

> it says it fixes #54 , but this has nothing to do with 54?

Fixed 😄 

> I don't see a default for the tag, which we should have

As we discussed and agreed on slack let's keep an empty string as default. I use the empty string to check if the feature has to be enabled or not. If not the handler is not registered at all.

> docs on how to configure it (README, mainly)
Coming soon!

packet/cloud.go
```diff
@@ -73,17 +74,22 @@ type Config struct {
...
 		instances:    newInstances(client, packetConfig.ProjectID),
 		zones:        newZones(client, packetConfig.ProjectID),
 		loadBalancer: newLoadBalancers(client, packetConfig.ProjectID, packetConfig.Facility, packetConfig.DisableLoadBalancer, packetConfig.LoadBalancerManifest, packetConfig.LocalASN, packetConfig.PeerASN),
 		bgp:          newBGP(client, packetConfig.ProjectID, packetConfig.LocalASN, packetConfig.PeerASN, packetConfig.AnnotationLocalASN, packetConfig.AnnotationPeerASNs, packetConfig.AnnotationPeerIPs),
-	}, nil
+	}
+	if packetConfig.EIPTag != "" {
+		c.controlPlaneEndpointManager = newControlPlaneEndpointManager(packetConfig.EIPTag, packetConfig.ProjectID, client.DeviceIPs, client.ProjectIPs, c.instances)
+	}
+	return c, nil
```
deitch, line 92: This here kind of breaks the style of the others. Is there a reason not to do it the same as the others

```go
return &cloud{
 		client:       client,
 		facility:     packetConfig.Facility,
 		instances:    newInstances(client, packetConfig.ProjectID),
 		zones:        newZones(client, packetConfig.ProjectID),
 		loadBalancer: newLoadBalancers(client, packetConfig.ProjectID, packetConfig.Facility, packetConfig.DisableLoadBalancer, packetConfig.LoadBalancerManifest, packetConfig.LocalASN, packetConfig.PeerASN),
 		bgp:          newBGP(client, packetConfig.ProjectID, packetConfig.LocalASN, packetConfig.PeerASN, packetConfig.AnnotationLocalASN, packetConfig.AnnotationPeerASNs, packetConfig.AnnotationPeerIPs),
                controlPlaneEndpointManager: newControlPlaneEndpointManager(packetConfig.EIPTag, packetConfig.ProjectID, client.DeviceIPs, client.ProjectIPs, c.instances),
 	}, nil
```

and then have `controlPlaneEndpointManager.nodeReconciler()` return nil if the first arg to `newControlPlaneEndpointManager()` is `""`?

packet/cloud.go
```diff
@@ -129,6 +135,11 @@ func (c *cloud) Initialize(clientBuilder cloudprovider.ControllerClientBuilder,
 			serviceReconcilers = append(serviceReconcilers, s)
 		}
 	}
+
+	if c.controlPlaneEndpointManager != nil {
+		nodeReconcilers = append(nodeReconcilers, c.controlPlaneEndpointManager.Reconciler())
+	}
+
```
deitch, line 142: Then this whole thing can go away.

deitch: And you can add `controlPlaneEndpointManager` to `services` slice, and it is all part of the same pattern.

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,131 @@
...
+	deviceIPSrv packngo.DeviceIPService
+	ipResSvr    packngo.ProjectIPService
+	projectID   string
+	httpClient  *http.Client
+}
+
+func (m *controlPlaneEndpointManager) Reconciler() func(nodes []*v1.Node, remove bool) error {
+	return func(nodes []*v1.Node, remove bool) error {
+		klog.V(2).Info("controlPlaneEndpoint.reconcile: new reconciliation")
+		ipList, _, err := m.ipResSvr.List(m.projectID)
```
deitch, line 47: Might want to add a safety check here that `m.eipTag` is not `""`.

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,131 @@
...
+		controlPlaneEndpoint := ipReservationByTags([]string{m.eipTag}, ipList)
+		if controlPlaneEndpoint == nil {
+			// IP NOT FOUND nothing to do here.
+			klog.Error("Elastic IP not found. Please verify you have one with the expected tag: %s", m.eipTag)
+			return err
+		}
+		klog.Infof("healthcheck elastic ip %s", controlPlaneEndpoint.Address)
+		req, err := http.NewRequest("GET", "https://"+controlPlaneEndpoint.Address+":6443/healthz", nil)
+		if err != nil {
+			klog.Errorf("healthcheck failed for control plane endpoint %s. err \"%s\". Checking the next one...", controlPlaneEndpoint.Address, err)
```
deitch, line 60: I don't understand this error. If you got an error here, it isn't the health check that failed, but the creation of a new request. And what does "checking the next one" mean? The next what?

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,131 @@
...
+	for _, node := range nodes {
+		addresses, err := m.instances.NodeAddresses(ctx, types.NodeName(node.Name))
+		if err != nil {
+			return err
+		}
+		// It does not matter if the IP is external or internal. We can reach both.
+		a := addresses[0]
+
+		klog.Infof("healthcheck node %s", a.Address)
+		req, err := http.NewRequest("GET", "https://"+a.Address+":6443/healthz", nil)
```
deitch, line 86: You might want a func to create the healthcheck URL. I can see us having a typo here at some point and not understanding why it still is broken, when we forgot to change it elsewhere.

Also, why would we hard-code `6443` as the port? That is the default, but it doesn't have to be that.

deitch: Sure. It definitely can default to `6443`, but it needs to be overridable.

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,131 @@
...
+			return err
+		}
+		controlPlaneEndpoint := ipReservationByTags([]string{m.eipTag}, ipList)
+		if controlPlaneEndpoint == nil {
+			// IP NOT FOUND nothing to do here.
+			klog.Error("Elastic IP not found. Please verify you have one with the expected tag: %s", m.eipTag)
+			return err
+		}
+		klog.Infof("healthcheck elastic ip %s", controlPlaneEndpoint.Address)
+		req, err := http.NewRequest("GET", "https://"+controlPlaneEndpoint.Address+":6443/healthz", nil)
```
displague, line 58: Should this use `http.NewRequestWithContext` with a time-out that can be configured externally?

displague: I see - https://github.com/packethost/packet-ccm/pull/58/files#diff-19330b406ae2e0189158188e75828febR121-R124

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+	}
+	return errors.New("ccm didn't find a good candidate for IP allocation. Cluster is unhealthy")
+}
+
+func newControlPlaneEndpointManager(eipTag, projectID string, deviceIPSrv packngo.DeviceIPService, ipResSvr packngo.ProjectIPService, i cloudInstances) *controlPlaneEndpointManager {
+	return &controlPlaneEndpointManager{
+		httpClient: &http.Client{
+			Timeout: time.Second * 5,
```
displague, line 121: I think we will want control over this timeout, perhaps as a command-line parameter.  Can we define this in a const for now?

packet/eip_controlplane_reconciliation.go
deitch: That is a good idea. @gianarb , how about we move the entire "make a health check" into its own func, which can be passed the URL? That way there is one place to control it, set time outs, etc.?

packet/eip_controlplane_reconciliation.go
deitch: We definitely want control. If you can get it in quickly @gianarb then great, but it is important we get this done before the weekend, so if it holds us up even a few mins, we can do a separate PR on it after.

deitch:
Also, this needs a rebase

packet/eip_controlplane_reconciliation.go
gianarb: In theory this is a good suggestion 

> Should this use http.NewRequestWithContext with a time-out that can be configured externally?

In practice, the reconciliation loop does not get any context from the outside so it won't be controllable as it is now

gianarb: @deitch the timeout is at the client level and I create the HTTP client for the all manager. I can change it but at the moment I don't see the win

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,167 @@
...
+		if err != nil {
+			return err
+		}
+		// It does not matter if the IP is external or internal. We can reach both.
+		a := addresses[0]
+
+		klog.Infof("healthcheck node %s", a.Address)
+		req, err := http.NewRequest("GET", fmt.Sprintf("https://%s:%d/healthz", a.Address, m.apiServerPort), nil)
+		if err != nil {
+			klog.Errorf("healthcheck failed for node %s. err \"%s\"", node.Name, err)
```
deitch, line 120: This error still is a bit odd. The healthcheck did _not_ fail, and doesn't unless the `httpClient.Do()` fails. This is a URL construction error.

packet/eip_controlplane_reconciliation.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (m *controlPlaneEndpointManager) init(k8sclient kubernetes.Interface) error {
+	m.k8sclient = k8sclient
+	klog.V(2).Info("controlPlaneEndpointManager.init(): enabling BGP on project")
+	return nil
+}
+
+func (m *controlPlaneEndpointManager) nodeReconciler() nodeReconciler {
+	return m.reconciler()
```
deitch, line 60: This is a bit convoluted. Why not just return it without invoking it?

```go
return m.reconciler
```

And then `reconciler()` can be:

```go
func (m *controlPlaneEndpointManager) reconciler(nodes []*v1.Node, remove bool) error {
    klog.V(2).Info("....")
    if m.inProcess {
// etc.
```

gianarb: `if m.inProcess {` it is not better to return as early as possible from the condition in order to avoid a level of nesting?
```
if !m.inProcess {
    return
}
```

?

deitch: yes, but I usually like to at least have the log, so when you run with `--v=2` you can know that it got there.

deitch:
This needs a rebase.

packet/eip_controlplane_reconciliation.go
deitch: The nesting shouldn't matter. This is a once a minute loop, very inexpensive for one more stack push-pop, the optimizer should handle it, and in any case, consistency is better for readability.

deitch:
Looks like it failed CI:

```
packet/cloud.go:78:2: struct field `ApiServerPort` should be `APIServerPort` (golint)
	ApiServerPort        int    `json:"apiServerPort,omitEmpty"`
	^
packet/constants.go:16:2: const `DefaultApiServerPort` should be `DefaultAPIServerPort` (golint)
	DefaultApiServerPort      = 6443
	^
main.go:34:2: const `envVarApiServerPort` should be `envVarAPIServerPort` (golint)
	envVarApiServerPort      = "PACKET_API_SERVER_PORT"
	^
make: *** [lint] Error 1
Makefile:120: recipe for target 'lint' failed
```

PR #60, opened 2020-07-09
add VERSION and apply this to the UserAgent

### LINKED ISSUE #54: Report UserAgent in API calls

displague:
_Originally from https://github.com/packethost/packet-ccm/pull/56#issuecomment-655528837_

This PR adds packet.VERSION and uses this in the packngo.UserAgent.  The Makefile was already applying the version at build time, but it was referencing a missing version variable (which we now have).

Closes #54


deitch:
Thanks, looks great!

PR #65, opened 2020-07-23
[WIP] for all peers, use node annotations instead of configmap

deitch:
**Do not merge**

This depends on an upstream PR from metallb, but implements the changes to absorb it. I will remove the `[WIP]` and the warning when this is ready.

Does the following:

* stops changing the configmap to update peers, using kubernetes node annotations instead
* adds source IP address to the annotations
* changes the annotation formatting slightly to match with how the metallb team does it (`-` instead of `.` as separator)
* fixes an error message to make it slightly clearer

Once the upstream PR is in, we will need to change the referenced image, and then this will be mergeable.


deitch:
This is long out of date. It will be replaced with a PR that focuses solely on getting the correct annotations.

PR #74, opened 2020-09-03
sending wrong plan name

### LINKED ISSUE #73: Set k8s built-in labels through the CCM
https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/

We should be setting some of those built-in labels like `node.kubernetes.io/instance-type` and `topology.kubernetes.io/zone` so that they match with packet naming.

deitch:
Fixes #73 

This fixes it. Before, it was

```yaml
    node.kubernetes.io/instance-type: baremetal_0
    topology.kubernetes.io/region: ewr1
```

with this, it is

```yaml
    node.kubernetes.io/instance-type: t1.small.x86
    topology.kubernetes.io/region: ewr1
```

### DIFF
--- a/packet/devices.go
+++ b/packet/devices.go
@@ -128,7 +128,7 @@ func (i *instances) InstanceType(_ context.Context, nodeName types.NodeName) (st
 		return "", err
 	}
 
-	return device.Plan.Slug, nil
+	return device.Plan.Name, nil
 }
 
 // InstanceTypeByProviderID returns the type of the specified instance.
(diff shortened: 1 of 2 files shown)


thebsdbox (dismissed):
The changes seem simple enough, is there an explanation for the ci failure?

deitch:
Yup, it had the wrong output for that func embedded in the tests too. Updated. Let's let I go clean, and we can merge it in.

thebsdbox (approved):
The CI light is green, this PR is clean!

`/lgtm`

PR #75, opened 2020-09-04
ensure CCM has access to configmaps

deitch:
Without access to configmaps, it could not set up metallb properly.

These access rights really should be restricted a little better, but for now, let's fix this real user's problem, and then worry about it.

### DIFF
--- a/deploy/template/deployment.yaml
+++ b/deploy/template/deployment.yaml
@@ -132,6 +132,17 @@ rules:
   - list
   - watch
   - update
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - create
+  - get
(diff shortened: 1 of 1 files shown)


thebsdbox (commented):
Given the error was only around the `get` of the `ConfigMap` do we want to allow all these additional actions/verbs?

thebsdbox:
ah ok, just re-read the first comment.

thebsdbox (approved):
This is OK for now to alleviate a customer issue, can we track this to tighten up permissions after the fact?

deitch:
Yeah, we need to do a better job here. 

It needs the ability to create, read, watch, update, patch the configmap.

PR #79, opened 2020-10-18
handle manifest errors correctly

deitch:
We were losing errors that occurred when trying to apply the metallb manifests. In a case where permissions were off (or anything else), nothing showed up in the logs.

This fixes that.

### DIFF
--- a/util/manifests.go
+++ b/util/manifests.go
@@ -30,6 +30,7 @@ func ApplyManifests(m []byte, client kubernetes.Interface) error {
 		return fmt.Errorf("error parsing manifests: %v", err)
 	}
 	for i, m := range manifests {
+		var err2 error
 		klog.V(2).Infof("applying manifest %d %v", i, m.GetObjectKind().GroupVersionKind())
 		// now figure out what kind it is
 		// for each kind, we get it. There are three possiblities:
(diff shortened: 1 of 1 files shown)


util/manifests.go
```diff
@@ -30,6 +30,7 @@ func ApplyManifests(m []byte, client kubernetes.Interface) error {
 		return fmt.Errorf("error parsing manifests: %v", err)
 	}
 	for i, m := range manifests {
+		var err2 error
```
jasmingacic, line 33: Wouldn't naming it manifestErr would make it a little more readable?

deitch: It might. I don't think it matters that much - everything in here is a manifest error, just a question of if it is an error from getting the resource or update/apply the resource.

thebsdbox: I suppose we never would really name the first `err` anything related to the types of errors it may hold.

thebsdbox (approved):
`/lgtm`

util/manifests.go
deitch: I guess we could rename the error inside each `case` statement as `getErr`, and then move `err2` back to `err`, or similar.

But does this really buy us anything?

Personally, I dislike the massive repetitiveness of the switch statement, but the `intf` in each `case` statement is something else, something that has `.Update(arg)` and `.Create(arg)` and `.Get(string, metva1.GetOptions)`, but the `arg` is different for each, so I found no way to abstract it all out.

PR #80, opened 2020-10-18
update rbac to reflect actual requirementS

deitch:
Also added lots and lots of comments on the ClusterRole, so we can have some audit trail of what it is there for.

### DIFF
--- a/deploy/template/deployment.yaml
+++ b/deploy/template/deployment.yaml
@@ -62,6 +62,14 @@ metadata:
   name: cloud-controller-manager
   namespace: kube-system
 ---
+# comments about this clusterrole
+# some of these are needed so that ccm can do its job
+# others are needed because metallb needs some roles; if ccm
+# tries to apply clusterrole or role permissions, and does not
+# have those permissions itself, it will fail when it attempts to
+# escalate privileges (as it should). All of those are in the lb/manifests.yaml
+#
+# for each one, we try to give the reasoning
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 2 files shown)


deploy/template/deployment.yaml
```diff
@@ -70,6 +78,57 @@ metadata:
   name: system:cloud-controller-manager
 rules:
 - apiGroups:
+  # reason: so ccm can create metallb-system namespace
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - create
+  - patch
```
thebsdbox, line 87: If we're only creating namespaces then I get the `list / get` to determine if they exist along with the `create` but do we need `patch`?

thebsdbox (dismissed):
Only really one comment that I don't understand, the rest looks kinda obvious for providing capability for the CCM to create service account etc.. given that the ccm runs in `cube-system` anyway it should be "relatively" guarded.

deploy/template/deployment.yaml
deitch: Probably not. The [code itself](https://github.com/packethost/packet-ccm/blob/master/util/manifests.go#L41-L48) is a similar loop for everything, even namespaces:

```go
		case *v1.Namespace:
			intf := client.CoreV1().Namespaces()
			existing, err := intf.Get(o.Name, metav1.GetOptions{})
			if err == nil && existing != nil {
				_, err2 = intf.Update(o)
			} else {
				_, err2 = intf.Create(o)
			}
```

If we removed the `intf.Update()` part, basically saying, "well, it exists, so I am happy", then we could get rid of the `patch` and `update`, at the cost of the code being slightly different for this `case` over every other.

I am ok either way - reduced perms with slightly more complex code vs simpler code with higher perms. What do you think @thebsdbox ?

thebsdbox: Ah if we're `updating` then I'm happy the way things are. :-)

thebsdbox:
`lgtm`

deitch:
I had to rebase. Do you mind re-checking and approve/comment @thebsdbox ?

thebsdbox (approved):
`/lgtm`

PR #82, opened 2020-10-20
add multiple peers

### LINKED ISSUE #63: Adjust CCM Load Balancer so that it adds all peering IPs returned by the Packet API / metadata
This #62 PR changed the way the CCM creates peers by pulling the BGP info through the Packet API. However, one drawback is that it only adds the first peering IP of the array returned by the API. 

https://github.com/packethost/packet-ccm/blob/30ebae601a916e3b620e4e547642b37bf9344c33/packet/loadbalancers.go#L140

Depending on the ToR router, the API will sometimes return more than 1 peering IP so we should loop through the array and add a peer config for each peering IP returned by the API.

deitch:
Fixes #63 

Adds all peers to the metallb configmap, and also to the annotations. Annotations are sorted and comma-separated for consistency.

### DIFF
--- a/packet/loadbalancers.go
+++ b/packet/loadbalancers.go
@@ -137,7 +137,7 @@ func (l *loadBalancers) reconcileNodes(nodes []*v1.Node, remove bool) error {
 				klog.Errorf("could not add metallb node peer address for node %s: %v", node.Name, err)
 				continue
 			}
-			config = addNodePeer(config, node.Name, peers[0], l.localASN, l.peerASN)
+			config = addNodePeer(config, node.Name, l.localASN, l.peerASN, peers...)
 			if config == nil {
 				klog.V(2).Info("config unchanged, not updating")
 				return nil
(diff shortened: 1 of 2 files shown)


packet/loadbalancers.go
```diff
@@ -274,19 +274,25 @@ func (l *loadBalancers) reconcileServices(svcs []*v1.Service, remove bool) error
 }
 
 // addNodePeer update the configmap to ensure that the given node has the given peer
-func addNodePeer(config *metallb.ConfigFile, nodeName string, peer string, localASN, peerASN int) *metallb.ConfigFile {
+func addNodePeer(config *metallb.ConfigFile, nodeName string, peers []string, localASN, peerASN int) *metallb.ConfigFile {
```
detiber, line 277: minor nit, but if changing the signature already, might make sense to make peers a variadic argument instead:
```suggestion
func addNodePeer(config *metallb.ConfigFile, nodeName string, localASN, peerASN int, peers ...string) *metallb.ConfigFile {
```

packet/loadbalancers.go
```diff
@@ -137,7 +137,7 @@ func (l *loadBalancers) reconcileNodes(nodes []*v1.Node, remove bool) error {
 				klog.Errorf("could not add metallb node peer address for node %s: %v", node.Name, err)
 				continue
 			}
-			config = addNodePeer(config, node.Name, peers[0], l.localASN, l.peerASN)
+			config = addNodePeer(config, node.Name, peers, l.localASN, l.peerASN)
```
detiber, line 140: It looks like we are assuming that all peers share the same ASN, is that intentional?

detiber (commented):
I'm not familiar with this code, so I'm not sure if the assumption of just a single peer ASN is expected or not. Outside of that and the minor nit these changes look good to me.

packet/loadbalancers.go
deitch: Yeah, I can do that. I thought about it and was lazy. I will do it now.

packet/loadbalancers.go
deitch: Yes. For now, that is what I have seen in every single use case; I never really have seen someone try to do something different. If it comes up (and Packet actually supports it), we can create that option then.

packet/loadbalancers.go
deitch: Updated.

deitch:
Updated. Please take another look.

detiber (approved):
lgtm

deitch:
Thanks @detiber 
I will merge once CI is green.

Issue #81, opened 2020-10-19
Reconsider MetalLB deployment management


displague:
The CCM's LoadBalancer feature, when enabled, creates and manages a MetalLB deployment. This includes the RBAC and service records.

While convenient, this managed installation expressing many opinions that are not configurable, including:
* namespace
* resource names
* RBAC settings
* all aspects of MetalLB configuration

Rather than managing the complete deployment, PacketCCM could offer an option to manage only the configmap used by a user-managed deployment.

* Remove deployment management (daemonset, services, rbac) for MetalLB resources from PacketCCM
* Remove RBAC packet-ccm uses in support of the controller management responsibility
* Keep ConfigMap management 
  * ConfigMap location should be a configuration option to the PacketCCM (PacketCCM's deployment would need to run with a clusterrole to access configmaps in any namespace)
  * ConfigMap type could be an option (to more easily support various versions of MetalLB or KubeVIP, assuming either can be managed by a single configmap)
  * ConfigMap management approach could be an option (patch vs create/replace, allowing users to provide base configuration that PacketCCM ignores)
* Optional Helm chart flags could be used to install MetalLB or KubeVIP
  * If we alternatively offer a kustomize based install, users could apply their own patches to the MetalLB yaml we provide
  * ^ may also be possible with Helm (I don't know Helm well enough to say)
  * Or these instructions could be left to the user, so the PacketCCM helm chart does not need to reproduce the helm chart of other applications. (the experience would be something like: helm install metallb; helm install packet-ccm --options-for-metallb)

---

Everything above still applies if we move to KubeVIP.

deitch:
To summarize what you and I discussed earlier today. Let's make sure we are on the same page.

Right now, ccm gives you two options:

1. Deploy metallb and manage the configmap for you ("enable")
1. Do not deploy metallb or do anything with it ("disable")

The first option is for those who want "it just works"; the second is for those who already know how to manage metallb and want to do it their own way.

You are proposing a third option: manage the config only.

Customers would deploy metallb if/how/wherever they want. ccm would be responsible solely for managing the configmap, as it already has the logic to do that. Customer would provide the config namespace and name, and ccm would manage it.

Is that correct?

displague:
@deitch, yes, that is how we discussed it.

I would perhaps, take it further, if possible.

I would take the "deploy metallb" responsibility out of the controller. MetalLB would be installed by Helm/Kustomize during the initial CCM install.

[...]

PR #83, opened 2020-10-22
normalize deployment of loadbalancer

deitch:
This is a pretty fundamental change.

Currently, we do the following with CCM inside the code:

1. enable BGP appropriately
1. read the metallb manifests from within the CCM container
1. deploy those manifests to kubernetes
1. modify the metallb configmap to update BGP peers and services. This is done on initialization and every add/delete of Kubernetes node or service

We have the option to disable the above using an env var and/or a setting in the secret.

This lead to several issues:

* complex deployment management, therefore complex code
* lots of `ClusterRole` settings for CCM just so it can pass those down to metallb
* undeploying CCM does not undeploy metalllb
* one of the biggest values in terms of loadbalancers is the last point: CCM does a great job managing the metallb configmap. Customers who want to deploy their own metallb but take advantage of CCM config management have no option; it is either fully on (CCM managed all the way) or disabled (no CCM benefit for load balancers)

This PR switches that by doing the following:

* CCM does not deploy metallb. However, along with each ccm `deployment.yaml` in Releases, it also provides a `loadbalancer.yaml` to deploy the load balancer. One can apply that, or not, as preferred
* When undeploying CCM, one also can undeploy metallb (if applied that way) by undeploying the `loadbalancer.yaml`
* CCM just manages the configmap, if it finds one. If not, it just ignores it.

It further can be configured to:

* use the default configmap `metallb-system:config`
* override to a different configmap using an env var and/or secret property
* disable entirely. This is useful if you are using the default `metallb-system:config`, but are managing it on your own and want CCM out of the way

This is intended directly to address #81 along with several other issues.

This needs more testing (always), but is ready for initial reviews.

Start with the README, as it describes the new and simpler flow.


deitch:
Fixes #81

displague:
It looks like `system:cloud-controller-manager` is missing create/update/patch roles on `service` and `endpoints`, needed by https://github.com/packethost/packet-ccm/blob/metallb-manifest/packet/eip_controlplane_reconciliation.go#L201-L319

README.md
```diff
@@ -173,34 +202,46 @@ For the control plane nodes, the Packet CCM uses static Elastic IP assignment, v
...
-  * Service deletion: find the `/32` IP address on the service spec and remove it; remove from the `ConfigMap`
-* Start an independent loop that checks every 30 seconds (configurable) for IP address reservations that are tagged with `usage="packet-ccm-auto"` and `service="<service-hash>"` but not on any services. If it finds one:
-  * If a service exists that matches the `<service-hash>`, that is an indication that an IP address reservation request was made, not completed at request time, and now is available. Add the IP to the [service spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#servicespec-v1-core) and ensure is in the pools of the metallb `ConfigMap` with `auto-assign: false`
-  * If no service exists that is missing an IP, or none with a matching hash, delete the IP reservation
+By default, CCM controls the loadbalancer by updating the `ConfigMap` named `metall-system:config`,
```
displague, line 205: `metallb-system` (missing b)

deitch: Nice catch. Updated, thanks.

deitch:
> It looks like system:cloud-controller-manager is missing create/update/patch roles on service and endpoints

Nice catch. Fixed up.

displague (approved):
This worked well for me. 

I deployed it using the cluster-api-provider-packet and ran into a snag because I deployed a single node cluster (the default configuration in the https://cluster-api.sigs.k8s.io/user/quick-start.html#initialize-the-management-cluster) and this single node included an untollerated `node-role.kubernetes.io/master` taint. Once I updated the metallb-controller to tolerate this, the sample service (and default/kubernetes) were both assigned EIPs and the services behind those IPs were accessible.

PR #87, opened 2020-11-02
handle odd returns from getting project BGP config

deitch:


packet/bgp.go
```diff
@@ -141,10 +141,18 @@ func (b *bgp) enableBGP() error {
...
+	// We treat it as valid config already exists only if ALL of the above is true:
+	// - no error
+	// - bgpConfig struct exists
+	// - bgpConfig struct has non-blank ID
+	// - bgpConfig struct does not have Status=="disabled"
+	if err == nil && bgpConfig != nil && bgpConfig.ID != "" && strings.ToLower(bgpConfig.Status) != "disabled" {
 		return nil
 	}
 
+	// we did not have a calid one, so created it
```
displague, line 155: typo here

image

displague: "so created it" .. and another one.

deitch: Haha. Nice catch. Fixed.

packet/bgp.go
```diff
@@ -141,10 +141,18 @@ func (b *bgp) enableBGP() error {
...
 	// if we already have a config, just return
-	if err == nil || bgpConfig != nil {
+	// we need some extra handling logic because the API always returns 200, even if
+	// not BGP config is in place.
+	// We treat it as valid config already exists only if ALL of the above is true:
+	// - no error
+	// - bgpConfig struct exists
+	// - bgpConfig struct has non-blank ID
+	// - bgpConfig struct does not have Status=="disabled"
+	if err == nil && bgpConfig != nil && bgpConfig.ID != "" && strings.ToLower(bgpConfig.Status) != "disabled" {
```
displague, line 151: Noting that the possible values for `bgpConfig.Status` are enabled, disabled, and requested.
This set of comparisons looks good.

displague (approved):

packet/bgp.go
displague: We may want to narrow down the range of errors we are ok with in the future (404s only?).

deitch: Agreed. We likely will have to redo this anyways once the BGP config endpoints are better structured.

PR #88, opened 2020-11-03
lots of logs on loadbalancer, plus return the error

deitch:
We had no logs at any level for when the loadbalancer reconciles nodes (adding them to the configmap). This adds those statements, mostly V(2) and V(5).

It also returns the error, which it should have done, but only return at the end of handling all of the nodes.

### DIFF
--- a/packet/loadbalancers.go
+++ b/packet/loadbalancers.go
@@ -100,46 +100,58 @@ func (l *loadBalancers) serviceReconciler() serviceReconciler {
 // by adding it to or removing it from the known metallb configmap
 func (l *loadBalancers) reconcileNodes(nodes []*v1.Node, remove bool) error {
 	var (
-		peers []string
-		err   error
+		peers        []string
+		err          error
+		changedNodes bool
+		changed      bool
 	)
+	klog.V(2).Infof("loadbalancers.reconcileNodes(): called for nodes %v", nodes)
 	// get the configmap
 	cmInterface := l.k8sclient.CoreV1().ConfigMaps(l.configmapnamespace)
 
+	klog.V(2).Infof("loadbalancers.reconcileNodes(): getting configmap %s/%s", l.configmapnamespace, l.configmapname)
 	config, err := getMetalConfigMap(cmInterface, l.configmapname)
 	if err != nil {
 		return fmt.Errorf("failed to get metallb config map %s:%s : %v", l.configmapnamespace, l.configmapname, err)
 	}
+	klog.V(5).Infof("loadbalancers.reconcileNodes(): original configmap %#v", config)
 	for _, node := range nodes {
 		// are we adding or removing the node?
 		if remove {
-			// go through the peers and see if we have one with our hostname.
-			selector := metallb.NodeSelector{
-				MatchLabels: map[string]string{
-					hostnameKey: node.Name,
-				},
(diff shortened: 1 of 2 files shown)


deitch:
More importantly, this fixes an issue wherein it could take minutes or longer for new nodes to get placed in the metallb configmap. This is pretty much immediate.

packet/loadbalancers.go
```diff
@@ -100,46 +100,58 @@ func (l *loadBalancers) serviceReconciler() serviceReconciler {
 // by adding it to or removing it from the known metallb configmap
 func (l *loadBalancers) reconcileNodes(nodes []*v1.Node, remove bool) error {
 	var (
-		peers []string
-		err   error
+		peers        []string
+		err          error
+		changedNodes bool
+		changed      bool
```
displague, line 106: I could be overlooking something, but it appears that changedNodes is only every set or used when changed is set, and in all cases it matches the value of changed.

packet/loadbalancers.go
```diff
@@ -100,46 +100,58 @@ func (l *loadBalancers) serviceReconciler() serviceReconciler {
...
+				klog.V(2).Infof("loadbalancers.reconcileNodes(): config changed for add %s", node.Name)
+				changedNodes = true
 			}
 		}
-		klog.V(2).Info("config changed, updating")
-		err = saveUpdatedConfigMap(cmInterface, l.configmapname, config)
 	}
-	return nil
+	if !changedNodes {
+		klog.V(2).Info("loadbalancers.reconcileNodes(): no changed to configmap, not updating")
```
displague, line 150: "no change"

displague (approved):
Looks good - nitpicks aside.

I'd really like to see some unit tests. The branches here are getting hard to hold in the head and I don't see the need for extensive mocking to provide tests here.

packet/loadbalancers.go
deitch: Nice catch. Fixing.

packet/loadbalancers.go
deitch: You did miss. The `changed` is set inside a loop for each node. `changedNodes` is the overall "have any nodes changed". In each loop (this is from the add):

```go
			config, changed = addNodePeer(config, node.Name, l.localASN, l.peerASN, peers...)
			if !changed {
				klog.V(2).Infof("loadbalancers.reconcileNodes(): config unchanged for add %s", node.Name)
			} else {
				klog.V(2).Infof("loadbalancers.reconcileNodes(): config changed for add %s", node.Name)
				changedNodes = true
			}
```

So `changed` captures "did this node change", which `changedNodes` is set to `true` if at least one of them had `changed`. We then use `changedNodes` to see if _any_ if the nodes have changed (thus requiring us to update the configmap).

deitch:
Updated the comment, so it needs a new review @displague . Thanks for taking the time on it.

deitch:
Also @displague I expect one more on Wednesday.

deitch:
> I'd really like to see some unit tests. The branches here are getting hard to hold in the head and I don't see the need for extensive mocking to provide tests here.

I agree, but not worth putting the effort in now, as so much is going to change. A lot of the headache we are dealing with now is metallb configmap-centric, which will go away soon enough. If it doesn't, we can worry about it then.

deitch:
> it needs a new review

Eh, you already approved, the only change was the log nit, so I am merging in.

PR #94, opened 2020-11-04
have mode to sync everything

### LINKED ISSUE #89: CCM Load Balancer doesn't remove node peers from the metalLB configmap when deleting nodes and missing the notification
The CCM has logic to remove node peers from the configmap, but only if it catches the nodes when they are being deleted. If it misses the delete node notification (like when CCM has to get restarted due to rescheduling from a deleted node to a healthy node), it has no provision for checking on each future loop (unlike adding, which does).

deitch:
Fixes #89 

We have an issue wherein this captures events as they happen, but if the ccm is down (e.g. pod restart or move), it won't actually do the correct deletes of nodes (from `peers`) or services (delete the EIP and remove from `pools` in configmap).

This fixes all of that by creating 3 modes for reconciliation functions:

* add: add whatever nodes/services it receives (idempotently, obviously)
* remove: remove whatever nodes/services it receives
* sync: take the provides list as authoritative, and make sure that the list of EIPs, peers and pools precisely matches that list

We already ran sync once per minute, but we didn't cover every use case.

### DIFF
--- a/packet/cloud.go
+++ b/packet/cloud.go
@@ -25,8 +25,8 @@ const (
 	checkLoopTimerSeconds        = 60
 )
 
-type nodeReconciler func(nodes []*v1.Node, remove bool) error
-type serviceReconciler func(services []*v1.Service, remove bool) error
+type nodeReconciler func(nodes []*v1.Node, mode UpdateMode) error
+type serviceReconciler func(services []*v1.Service, mode UpdateMode) error
 
 // cloudService an internal service that can be initialize and report a name
 type cloudService interface {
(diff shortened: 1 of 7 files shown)


deitch:
This is ready for review @displague

packet/ipreservations_test.go
```diff
@@ -26,7 +26,7 @@ func TestIPReservationByTags(t *testing.T) {
 	}
```
displague, line 26: A case with multiple and no tags would be helpful, since this is testing AllTags now. 
```
		{[]string{}, 0},
		{[]string{"b", "c"}, 3},
```

The name of the test function should be updated too (to `AllTags`).

You may want to copy/paste this as is to test `AnyTags`.

displague: I'll follow this PR up with a linter/coverage PR: https://github.com/packethost/packet-ccm/compare/master...displague:lint-and-cover?expand=1

deitch: Yeah, that is a good idea.

deitch: And doing so caught an error. All fixed now.

deitch:
Updated @displague ; take another look.

packet/ipreservations_test.go
displague: The `AllTags` test is not testing for cases where multiple tags were expected to match (all of the cases have a single tag).

Neither test is including empty cases (both functions would currently return a 0 index match).

deitch: Good idea. Got it now. Take another look @displague

displague (approved):

packet/cloud.go
```diff
@@ -176,47 +176,41 @@ func (c *cloud) HasClusterID() bool {
...
+				if err := h([]*v1.Node{n}, ModeRemove); err != nil {
 					klog.Errorf("failed to update and sync node for remove %s for handler: %v", n.Name, err)
 				}
 			}
 		},
 	})
+
 	klog.V(5).Info("startNodesWatcher(): nodesInformer.Run()")
 	go nodesInformer.Run(stop)
+	syncFuncs := []cache.InformerSynced{
```
displague, line 207: I had to read up on this to figure out what was going on here.
A comment could help future reviews.

https://engineering.bitnami.com/articles/a-deep-dive-into-kubernetes-controllers.html

deitch: Definitely. I will open a new PR with better documentation.

FWIW, _every time_ I go near this, I need to read up again. 😂

deitch: See #97 @displague , just for you. :-)

PR #95, opened 2020-11-04
remove strict parsing

### LINKED ISSUE #93: re-applying Lokomotive's MetalLB fails ccm
### Steps to reproduce

- Apply secrets
- Apply [deployment](https://github.com/packethost/packet-ccm/blob/master/deploy/template/deployment.yaml) on cluster with kubelet configured.
- Apply [Lokomotive MetalLb](https://github.com/kinvolk/lokomotive/blob/master/docs/configuration-reference/components/metallb.md) component. 

If component is re-applied, ccm does not pick configMap correctly and throws error that configMap is not correct, which should not happen. You have to delete the pod manually to get it to work again.

```sh
failed to update and sync nodes: failed to get metallb config map: could not parse config: yaml: unmarshal errors:
  line 1: field peer-autodiscovery not found in type metallb.ConfigFile
```

cc @deitch

deitch:
Fixes #93

### DIFF
--- a/packet/metallb/configmap.go
+++ b/packet/metallb/configmap.go
@@ -14,7 +14,7 @@ import (
 
 func ParseConfig(bs []byte) (*ConfigFile, error) {
 	var raw ConfigFile
-	if err := yaml.UnmarshalStrict(bs, &raw); err != nil {
+	if err := yaml.Unmarshal(bs, &raw); err != nil {
 		return nil, fmt.Errorf("could not parse config: %s", err)
 	}
 


displague (approved):
lgtm.

I'm curious about which fields were not mapped or were duplicated? Should they be better handled or are they extraneous to our needs?

deitch:
> I'm curious about which fields were not mapped or were duplicated? Should they be better handled or are they extraneous to our needs?

Extraneous to our needs. Some of the Kinvolk folk extend the configmap. No reason we should break on that, we can be expansive 😀

PR #97, opened 2020-11-08
update documentation on controller lifecycle

deitch:
Just to make things easier for future reviewers.

### DIFF
--- a/packet/cloud.go
+++ b/packet/cloud.go
@@ -202,6 +202,19 @@ func startNodesWatcher(informer informers.SharedInformerFactory, handlers []node
 		},
 	})
 
+	// what this does:
+	// when you create an informer, you start it by calling informer.Run()
+	// however, it can take some time for the local state to sync up. If you use any methods before
+	// it is completely synced, especially get or list, you can end up missing data. In order to
+	// avoid the issue, you run it in the following order:
+	//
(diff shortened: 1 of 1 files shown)


packet/cloud.go
```diff
@@ -202,6 +202,19 @@ func startNodesWatcher(informer informers.SharedInformerFactory, handlers []node
...
 
+	// what this does:
+	// when you create an informer, you start it by calling informer.Run()
+	// however, it can take some time for the local state to sync up. If you use any methods before
+	// it is completely synced, especially get or list, you can end up missing data. In order to
+	// avoid the issue, you run it in the following order:
+	//
+	// 1. create your informer
+	// 2. informer.Run()
+	// 3. create a slice of sync functionse []cache.InformerSynced. The function on each informer is informer.HasSynced
```
displague, line 213: functionse

displague (approved):
LGTM - In the future, maybe we can refactor the two uses into one function and simply use these notes as documentation for that function.

packet/cloud.go
deitch: Yeah. In case gollum reads it. "functionse"!

deitch: Fixed.

PR #98, opened 2020-11-09
manifest-tool was failing with missing binder

deitch:
Oops

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,8 +166,11 @@ ifneq (,$(DESTDIR))
 	cp $(DIST_BINARY) $(DESTDIR)/$(shell basename $(DIST_BINARY))
 endif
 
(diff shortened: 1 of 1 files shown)


deitch:
We really should clean up this Makefile to move a lot of it into actions. Not for today, though.

displague (approved):
This seems ok. I would like to know under what conditions a GOPATH exists without a GOBIN. I'm wondering if we end up installing to `/bin` in this case.

deitch:
>  I would like to know under what conditions a GOPATH exists without a GOBIN. I'm wondering if we end up installing to /bin in this case.

Not sure. I think it has to do with actions. Either way, as long as it works, I will be happy. Of course, I cannot test without cutting a release, but so be it.

PR #101, opened 2020-11-10
Add support for BGP node selector

### LINKED ISSUE #100: Add ability to select on which nodes BGP will be enabled
In some large clusters, one may want to have dedicated pool of Ingress nodes, which will be handling ingress traffic. CCM currently enables BGP on all nodes in the cluster.

In [Lokomotive](https://github.com/kinvolk/lokomotive) we allow users to disable BGP per-worker pool, if needed, for example for security reasons.

It would be great if one could configure a node selector in the CCM, which would select on which nodes BGP should be enabled.

invidian:
In some large clusters, one may want to have dedicated pool of Ingress
nodes, which will be handling ingress traffic. CCM currently enables BGP
on all nodes in the cluster.

Enabling BGP on all nodes in the cluster also has security implications,
as now every node can start running it's own BGP speaker and start
capturing incoming traffic.

This patch adds support for new configuration option called BGP Node
Selector, which allows specifying Kubernetes Label selector, which will
be used for filtering the nodes, which will have BGP enabled.

To preserve existing behavior, by default all Nodes are selected.

Closes #100

### DIFF
--- a/main.go
+++ b/main.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"time"
 
+	"k8s.io/apimachinery/pkg/labels"
 	cliflag "k8s.io/component-base/cli/flag"
 	"k8s.io/component-base/logs"
 	_ "k8s.io/component-base/metrics/prometheus/clientgo" // for client metric registration
@@ -33,6 +34,7 @@ const (
 	envVarAnnotationPeerIPs      = "PACKET_ANNOTATION_PEER_IPS"
 	envVarEIPTag                 = "PACKET_EIP_TAG"
 	envVarAPIServerPort          = "PACKET_API_SERVER_PORT"
+	envVarBGPNodeSelector        = "PACKET_BGP_NODE_SELECTOR"
 	defaultLoadBalancerConfigMap = "metallb-system:config"
 )
 
@@ -209,6 +211,15 @@ func getPacketConfig(providerConfig string) (packet.Config, error) {
 		config.APIServerPort = rawConfig.APIServerPort
 	default:
 		config.APIServerPort = packet.DefaultAPIServerPort
+	}
+
+	config.BGPNodeSelector = rawConfig.BGPNodeSelector
+	if v := os.Getenv(envVarBGPNodeSelector); v != "" {
+		config.BGPNodeSelector = v
+	}
+
+	if _, err := labels.Parse(config.BGPNodeSelector); err != nil {
+		return config, fmt.Errorf("BGP Node Selector must be valid Kubernetes selector: %w", err)
 	}
 
 	return config, nil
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -211,6 +213,15 @@ func getPacketConfig(providerConfig string) (packet.Config, error) {
 		config.APIServerPort = packet.DefaultAPIServerPort
 	}
 
+	config.BGPNodeSelector = rawConfig.BGPNodeSelector
+	if v := os.Getenv(envVarBGPNodeSelector); v != "" {
+		config.BGPNodeSelector = v
+	}
```
deitch, line 219: Can we have a CLI flag option as well, like most of the ones above?

packet/bgp.go
```diff
@@ -22,9 +23,16 @@ type bgp struct {
 	k8sclient                                                 kubernetes.Interface
 	peerASN, localASN                                         int
 	annotationLocalASN, annotationPeerASNs, annotationPeerIPs string
+	nodeSelector                                              labels.Selector
 }
 
-func newBGP(client *packngo.Client, project string, localASN, peerASN int, annotationLocalASN, annotationPeerASNs, annotationPeerIPs string) *bgp {
+func newBGP(client *packngo.Client, project string, localASN, peerASN int, annotationLocalASN, annotationPeerASNs, annotationPeerIPs string, nodeSelector string) *bgp {
```
deitch, line 29: I think it is high time this got moved into a parameter struct, or variadic function parameters. But not for this PR; one issue at a time.

deitch (changes requested):
This looks quite nice. Just a small addition. When you think it is past draft status, I can ask some EM people to test it.

main.go
invidian: Can you elaborate how to add it? I cannot figure it out from the code. Passing `-h` to the binary is not doing anything either, so I cannot even check what flags are available. README.md doesn't mention any CLI flag other than `-v` either :confused:

deitch: Huh. Never mind, I thought we had added all of that here, but it is in a different project. Ignore this.

invidian:
>This looks quite nice. Just a small addition. When you think it is past draft status, I can ask some EM people to test it.

I'll test it myself first to make sure it works as I expect and then mark as ready for reviews, thanks!

packet/bgp.go
invidian: Created #102 for tracking then.

invidian:
Okay, I've tested it and it works as expected. You can see how this is used in https://github.com/kinvolk/lokomotive/pull/1155.

invidian:
Ping @deitch

deitch:
Cleaning up some open issues, this looks good @invidian . Do you want to rebase and we can get it in?

invidian:
On it.

invidian:
Rebased.

deitch (approved):

PR #108, opened 2020-11-13
deploy/template/deployment.yaml: remove autoupdate annotation

invidian:
According to Kubernetes documentation, autoupdate is true by default
when RBAC mode is enabled, so this annotation is redundant and can be
removed.

### DIFF
--- a/deploy/template/deployment.yaml
+++ b/deploy/template/deployment.yaml
@@ -65,8 +65,6 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  annotations:
-    rbac.authorization.kubernetes.io/autoupdate: "true"
   name: system:cloud-controller-manager
 rules:
 - apiGroups:


invidian:
Ping @deitch

displague:
sgtm https://kubernetes.io/docs/reference/access-authn-authz/rbac/#auto-reconciliation

deitch:
If this is a default role with default permissions, why would we need to define `ClusterRole system:cloud-controller-manager` at all? Is it possible that this is required when you modify the default role? I don't actually know.

invidian:
>If this is a default role with default permissions, why would we need to define ClusterRole system:cloud-controller-manager at all? Is it possible that this is required when you modify the default role? I don't actually know.

Hm, I don't see `system:cloud-controller-manager` ClusterRole being automatically created by `kube-apiserver` on my cluster, so I think it still needs to be defined here.

displague:
On second think, why are we manipulating `system:` clusterroles? Perhaps we should create a custom clusterrolebinding that extends system:cloud-controller-manager, specific to the equinix-metal ccm.

Other CCMs don't create or update this clusterrole.

invidian:
@displague and where is `system:cloud-controller-manager` originally defined? As I said, I cannot see it on my cluster without the CCM...

displague:
DO does something very similar to what we are doing: https://github.com/digitalocean/digitalocean-cloud-controller-manager/blob/master/releases/v0.1.30.yml#L58-L143

displague:
@invidian I see. This appears to be a convention, not something preinstalled.

I was fooled by https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#examples which creates a clusterrolebinding without defining the clusterrole.

invidian:
I'm curious what is the "right way" of setting these things up.

deitch:
I honestly cannot recall the history of this. CCM has gone through more than a few iterations, and was much harder to set up earlier on. I don't actually remember the precise source of that - I do remember working with some of the DO folks, although not specifically on this - but I do remember it being something of a convention. I would be hesitant to change its name from `system:` anything unless and until we know for sure that it won't run afoul of something else.

I _think_ it is the "moral equivalent" of putting it in `namespace: kube-system`, but of course `ClusterRole`s are not in namespaces.

deitch:
In any case, is there a specific reason _not_ to leave the `autoupdate` annotation?

invidian:
>In any case, is there a specific reason not to leave the autoupdate annotation?

Well, as this is the Kubernetes default, having it defined explicitly is redundant and this leads to more code to maintain etc. I think if you cannot say a specific reason why something is written (or have tests for it to prove that it is required), then it should probably be removed.

In any case, if "keeping the legacy" is such a reason, I'm fine with closing this PR. Though perhaps some comment/issue to clean this up at some point in the future would be nice.

deitch:
> Well, as this is the Kubernetes default, having it defined explicitly is redundant and this leads to more code to maintain etc. I think if you cannot say a specific reason why something is written (or have tests for it to prove that it is required), then it should probably be removed

Given that it is Kubernetes, I am wary of removing anything that might break something else, when it doesn't hurt, also known as:

> if "keeping the legacy" is such a reason

:-)

displague:
Perhaps there is less worry about introducing this change, now that we are at a new image, product name, and version (would have been nice to get it in at 3.0.0 though).

Issue #113, opened 2020-12-03
Update Kubernetes dependencies to fix printing "unable to load initial CA bundle for" warning


invidian:
Specifically to include https://github.com/kubernetes/kubernetes/commit/413960e49bea4b5558ea4dda3d18137eceaf7f16#diff-cede7efa8beef64baa0651bd73b39e27ad9508763719b48b4d88483a61964c89, as currently following warning is being printed for no reason:
```
W1203 10:38:04.249751       1 configmap_cafile_content.go:102] unable to load initial CA bundle for: "client-ca::kube-system::extension-apiserver-authentication::client-ca-file" due to: configmap "extension-apiserver-authentication" not found
W1203 10:38:04.249802       1 configmap_cafile_content.go:102] unable to load initial CA bundle for: "client-ca::kube-system::extension-apiserver-authentication::requestheader-client-ca-file" due to: configmap "extension-apiserver-authentication" not found
```

detiber:
I'm currently attempting to update the dependencies, wish me luck :joy:

invidian:
Good luck! :smile: Maybe you find https://github.com/kubernetes-sigs/clientgofix useful? I never used it, but found it some time ago.

PR #119, opened 2020-12-03
Update dependencies

detiber:
- Update kubernetes dependencies to v1.19.4
- Update packngo to v0.5.1
- Update go to v1.15.6

Fixes: #113


deitch (approved):
I wouldn't hold it up for this, but what do we gain by passing a `ctx.Done()` (which returns a `<- chan struct{}`) to each `nodesInformer.Run()`, as opposed to the existing `stop <- chan struct{}` ? Is it for fan-out?

detiber:
> I wouldn't hold it up for this, but what do we gain by passing a `ctx.Done()` (which returns a `<- chan struct{}`) to each `nodesInformer.Run()`, as opposed to the existing `stop <- chan struct{}` ? Is it for fan-out?

It was mainly to facilitate the plumbing of context through and to avoid creating multiple contexts (in each reconcile method) or having to pass both a context and a channel to the methods that needed them.

Issue #117, opened 2020-12-03
LoadBalancer enabling behaviours - what is the correct UX


deitch:
Currently, the ccm has an option to manage loadbalancers, specifically metallb.

The ccm takes the job of:

1. interfacing with the Packet API to get an EIP to assign to to a `Service type=LoadBalancer`
1. attaching that EIP to the `Service`, specifically `Spec.Status.LoadBalancerIP`
1. interfacing with the Packet API to enable BGP on the project and per-device
1. placing BGP information on each node's annotations

Note that none of the above has anything to do with a specific implementation (metallb). Packet is unique in that it could have multiple different loadbalancer implementations. Unlike AWS (ELB) or other providers, it isn't tightly tied to one.

In addition, because the (current) LB implementation _is_ metallb, it also does:

5. manage the metallb configmap by placing the bgp and service information in it

Since the two - general lb requirements (list of first 4); metallb-specific (last 1) - have been closely intertwined, ccm has used the existence of the configmap as a "switch" to determine whether or not to do the rest. The UI looks something like this. It is controlled via the value of the configmap flag, env var `PACKET_LB_CONFIGMAP` or equivalent.

* `disabled` - do nothing, do not try to manage the metallb configmap (item 5), but also don't do any non-metallb stuff (items 1-4)
* `` (empty) - default to the metallb default, `metallb-system:config`
* `<something:else>` - use that as the config map namespace+name

If the field is not `disabled` and not empty, then:

* if the `configmap` exists, manage that configmap (item 5) and the other non-metallb stuff (items 1-4)
* if the `configmap` does not exist, do not manage any configmap (item 5) or any of the other non-metallb stuff (items 1-4)

My contention is that we should be able to separate the two lists of metallb-specific and non-metallb-specific, and have a choice to:

* disable everything - we have this now with `PACKET_LB_CONFIGMAP`, which is a bad name for this
* enable just the first list (items 1-4)  - does not exist
* enable the first list (items 1-4) and the second (item 5) - we have this now if the configmap exists and the env var is not `disabled`

Question: what should this configuration look like. How should we tell ccm to pick option a, b or c? Obviously env vars / configs / cli flags, as usual, but what specifically should they be and what should the options be set to?

Let's keep in mind that we may support other LB implementations in the future.

If we can come up with a sane UI with the right flexibility, then we can implement it quickly.

cc @thebsdbox @gianarb @displague @detiber @jmarhee

thebsdbox:
[...]

PR #121, opened 2020-12-04
Configmap is succesfully ignored and status updated

thebsdbox:
This is the initial commit to move to a CCM that we can divert the code path away from `metallb` and ensure that the `status` is correctly updated for "general purpose" Loadbalancers.

### DIFF
--- a/main.go
+++ b/main.go
@@ -115,8 +115,13 @@ func getPacketConfig(providerConfig string) (packet.Config, error) {
 	if config.LoadBalancerConfigMap == "" {
 		config.LoadBalancerConfigMap = defaultLoadBalancerConfigMap
 	}
+
 	if config.LoadBalancerConfigMap == "disabled" {
 		config.LoadBalancerConfigMap = ""
+	}
+
+	if loadBalancerConfigMap == "none" {
+		config.LoadBalancerConfigMap = loadBalancerConfigMap
 	}
 
 	facility := os.Getenv(facilityName)
--- a/main.go
+++ b/main.go
@@ -107,21 +107,25 @@ func getPacketConfig(providerConfig string) (packet.Config, error) {
 
 	loadBalancerConfigMap := os.Getenv(loadBalancerConfigMapName)
 	config.LoadBalancerConfigMap = rawConfig.LoadBalancerConfigMap
+
+	// Configmap behaviour:
+	// If it isn't blank, set it to the input
+	// If it is blank, set it to the default
+	// If it isn't blank but is "disabled" then disable cconfigmap watching
+
 	// rule for processing: any setting in env var overrides setting from file
 	if loadBalancerConfigMap != "" {
 		config.LoadBalancerConfigMap = loadBalancerConfigMap
+		config.ConfigMapEnable = true
 	}
 	// and set for default
 	if config.LoadBalancerConfigMap == "" {
 		config.LoadBalancerConfigMap = defaultLoadBalancerConfigMap
+		config.ConfigMapEnable = true
 	}
-
 	if config.LoadBalancerConfigMap == "disabled" {
-		config.LoadBalancerConfigMap = ""
-	}
-
-	if loadBalancerConfigMap == "none" {
-		config.LoadBalancerConfigMap = loadBalancerConfigMap
+		config.LoadBalancerConfigMap = "disabled"
+		config.ConfigMapEnable = false
 	}
 
 	facility := os.Getenv(facilityName)
--- a/packet/config.go
+++ b/packet/config.go
@@ -8,6 +8,7 @@ type Config struct {
 	ProjectID             string  `json:"projectId"`
 	BaseURL               *string `json:"base-url,omitempty"`
 	LoadBalancerConfigMap string  `json:"loadbalancer-configmap"`
+	ConfigMapEnable       bool    `json:"configmap-disable"`
 	Facility              string  `json:"facility,omitempty"`
 	PeerASN               int     `json:"peerASN,omitempty"`
 	LocalASN              int     `json:"localASN,omitempty"`
--- a/packet/loadbalancers.go
+++ b/packet/loadbalancers.go
@@ -30,35 +30,33 @@ type loadBalancers struct {
 	project            string
 	configmapnamespace string
 	configmapname      string
+	configmapenable    bool
 	facility           string
 	localASN, peerASN  int
 }
 
-func newLoadBalancers(client *packngo.Client, projectID, facility string, configmap string, localASN, peerASN int) *loadBalancers {
+func newLoadBalancers(client *packngo.Client, projectID, facility, configmap string, configmapenable bool, localASN, peerASN int) *loadBalancers {
 	// parse the configmap - we only accept a full namespace and name, no fallback defaults
 	var configmapnamespace, configmapname string
 	cmparts := strings.SplitN(configmap, ":", 2)
 	if len(cmparts) >= 2 {
 		configmapnamespace, configmapname = cmparts[0], cmparts[1]
 	}
-	if configmap == "none" {
-		configmapname = configmap
-	}
-	return &loadBalancers{client, nil, projectID, configmapnamespace, configmapname, facility, localASN, peerASN}
+	return &loadBalancers{client, nil, projectID, configmapnamespace, configmapname, configmapenable, facility, localASN, peerASN}
 }
 
 func (l *loadBalancers) name() string {
 	return "loadbalancer"
 }
+
 func (l *loadBalancers) init(k8sclient kubernetes.Interface) error {
 	klog.V(2).Info("loadBalancers.init(): started")
 	l.k8sclient = k8sclient
 
-	if l.configmapname == "" {
-		klog.V(2).Info("loadBalancers disabled, not managing metallb")
-		return nil
+	if !l.configmapenable {
+		klog.V(2).Info("No ConfigMap has been entered to watch, not managing metallb")
 	}
-	// deploy metallb
+
 	klog.V(2).Info("loadBalancers.init(): complete")
 	return nil
 }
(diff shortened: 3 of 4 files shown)


packet/loadbalancers.go
```diff
@@ -63,6 +67,18 @@ func (l *loadBalancers) init(k8sclient kubernetes.Interface) error {
 // we do this via metallb, not directly, so none of this works... for now.
 
 func (l *loadBalancers) GetLoadBalancer(ctx context.Context, clusterName string, service *v1.Service) (status *v1.LoadBalancerStatus, exists bool, err error) {
+	// If a service has had a loadbalancer address applied via the alternative reconcilliation loop then we can return it as the
```
thebsdbox, line 70: This can actually be removed now (the status is updated in the other reconciliation)

main.go
```diff
@@ -115,10 +115,15 @@ func getPacketConfig(providerConfig string) (packet.Config, error) {
 	if config.LoadBalancerConfigMap == "" {
 		config.LoadBalancerConfigMap = defaultLoadBalancerConfigMap
 	}
+
 	if config.LoadBalancerConfigMap == "disabled" {
 		config.LoadBalancerConfigMap = ""
 	}
 
+	if loadBalancerConfigMap == "none" {
```
deitch, line 123: I don't think this is necessary, as it already is covered in L115-117 above.

packet/loadbalancers.go
```diff
@@ -41,6 +41,9 @@ func newLoadBalancers(client *packngo.Client, projectID, facility string, config
 	if len(cmparts) >= 2 {
 		configmapnamespace, configmapname = cmparts[0], cmparts[1]
 	}
+	if configmap == "none" {
+		configmapname = configmap
+	}
```
deitch, line 46: So this now sets the `configmapnamespace="none"` and `configmapname="none"`?

packet/loadbalancers.go
deitch: I don't understand the comment. If we can remove this, why are we adding it?

main.go
thebsdbox, line 123: This will be removed and we'll actually pass "disabled" in as the Configmap name and not "none", which will be parsed in the `loadbalancer.go`

packet/loadbalancers.go
deitch: More importantly, I have 2 questions:

1. are we sure that these do not conflict? Is the "right" way to set the `LoadBalancerIP` here via the `Status` and not there via the `Spec.LoadBalancerIP`?
1. Are we sure that setting the status does not mess up metallb-system? We need to ensure that keeps working correctly. It is worth digging into how the speaker there works, to make sure that it is not expecting to find the IP and set the status, and perhaps ignore it if the status is set already?

packet/loadbalancers.go
```diff
@@ -412,11 +434,18 @@ func (l *loadBalancers) addService(svc *v1.Service, ips []packngo.IPAddressReser
...
 		existing.Spec.LoadBalancerIP = svcIP
 
-		_, err = intf.Update(existing)
+		updatedService, err := intf.Update(existing)
 		if err != nil {
 			klog.V(2).Infof("failed to update service %s: %v", svcName, err)
 			return fmt.Errorf("failed to update service %s: %v", svcName, err)
 		}
+
+		updatedService.Status.LoadBalancer.Ingress = []v1.LoadBalancerIngress{{IP: svcIP}}
```
deitch, line 443: Hmm, this appears duplicate of above?

packet/loadbalancers.go
```diff
@@ -41,6 +41,9 @@ func newLoadBalancers(client *packngo.Client, projectID, facility string, config
 	if len(cmparts) >= 2 {
 		configmapnamespace, configmapname = cmparts[0], cmparts[1]
 	}
+	if configmap == "none" {
```
thebsdbox, line 44: Our "none" or "disabled" doesn't follow the "namespace:configmapname" format, so we need to persist it after split above so that the loadbalancer reconciliation loop knows what to do.

packet/loadbalancers.go
deitch: So is `"none"` a way to say, "do all of your loadbalancer stuff, but do not do metallb stuff"? Now we have 2 modes:

* `disabled` - do not do LB
* not-`disabled` (either `""` which defaults to `"metallb-system:config"`, or something else explicit) - do LB and metallb config

This would add a third mode:

* `none` - do all of the LB stuff _except_ for the metallb configmap

Is that it?

deitch (commented):
Not quite ready to sign off, as I had a few questions.

If I did understand correctly, I would like to see us resolve #117 in a general sense before merging this one in.

We also need to be really sure that we do not break metallb.

deitch:
Hmm, looking [here](https://github.com/metallb/metallb/blob/5fff8b37444fc11b4bce3a6c2ac6cfedb479e4a2/controller/service.go#L27-L137), especially the end, it looks like the controller for metallb does set the same setting as this PR does. 

It looks from [here](https://github.com/metallb/metallb/blob/5fff8b37444fc11b4bce3a6c2ac6cfedb479e4a2/controller/service.go#L49-L54) that it actually tries to read it first. So I don't _think_ what you are doing will conflict, but I am unsure. This would need some testing.

packet/loadbalancers.go
deitch: Got it. I actually find this confusing. We use the `disabled` as a user-facing flag, but at the entry point (not in `loadbalancers.go`), resolve it to the default name.

If this is meant to indicate, "don't use the configmap, but keep doing bgp stuff", then we probably should have a boolean flag for that, rather than overloading it. It makes sense to overload for the env var, as that is how the user indicates intent, but not so much for a value passed to a func, as there we have options.

packet/loadbalancers.go
thebsdbox: The speaker does exactly the same thing. Once it has done its work it will also update the status. The code above was left in from my work last night. Didn’t realise I’d left it in when I commited this morning.

packet/loadbalancers.go
thebsdbox: Happy with a cool bool.

packet/loadbalancers.go
thebsdbox: 1. yes.. the status is reflect the current "state" of the load balancer.. the spec reflects the "expected state" of the load balancer

2. metallb will do the same thing, once  it does it's thing it will do an `updateStatus(sevicev1)`

main.go
thebsdbox: This will be removed and we'll transition to a bool in the load balancer to disable all ConfigMap calls.

packet/loadbalancers.go
deitch: 1. So we need both? What is the "right place" to set the `Spec` and what is the "right place" to set the `Status`?
1. for metallb, see my comments below.

thebsdbox: 1. We set both in the original place we set the `Spec` as mentioned first.. this code was put in the wrong place by me in testing.

packet/loadbalancers.go
thebsdbox: This is the correct place.. the GetLoadBalancer() is never called.

thebsdbox:
After talking with @displague it's clear I took the wrong architecture choice.

The CCM has the responsibility to update the spec. However it doesn't  implement the functionality. So I've removed the Status update and kube-vip or metallb will do that once they "implement" the networking.

thebsdbox:
So my scope of changes now are really about deferring the ConfigMap functionality.

packet/config.go
```diff
@@ -8,6 +8,7 @@ type Config struct {
 	ProjectID             string  `json:"projectId"`
 	BaseURL               *string `json:"base-url,omitempty"`
 	LoadBalancerConfigMap string  `json:"loadbalancer-configmap"`
+	ConfigMapEnable       bool    `json:"configmap-disable"`
```
deitch, line 11: The field of the struct is `ConfigMapEnable`, but the json is `"configmap-disable"`?

deitch:
@thebsdbox can you describe what the UX is, then? I _think_ but still not totally clear, that with this PR, I will have several options:

* disabled entirely
* enabled for basic BGP stuff (in my other issue, items 1-4) but not metallb-configmap control (item 5)
* enabled for basic BGP stuff (items 1-4) and metallb-configmap control (item 5)

What are the settings for the user to set all of these? It might be easiest if you update the PR to include changes to the README to describe it.

packet/loadbalancers.go
```diff
@@ -30,31 +30,33 @@ type loadBalancers struct {
 	project            string
 	configmapnamespace string
 	configmapname      string
+	configmapenable    bool
 	facility           string
 	localASN, peerASN  int
 }
 
-func newLoadBalancers(client *packngo.Client, projectID, facility string, configmap string, localASN, peerASN int) *loadBalancers {
+func newLoadBalancers(client *packngo.Client, projectID, facility, configmap string, configmapenable bool, localASN, peerASN int) *loadBalancers {
```
deitch, line 38: This is a bit of a strange interface. I have the `configmap` as a string, which I use both to give the name of the configmap and whether it is enabled, and then a `configmapenable` which determines if it is enabled (didn't we just do that?).

It would be good if we could pull the complexity (largely driven by the user-facing settings) out of here entirely, and made this interface easier to understand. Maybe something like these parameters:

* `configmap string` - if blank, do not do configmap stuff; if non-blank, it is the namespace+name of the configmap
* `enabled bool` - if false, do not do any loadbalancer stuff; if true, enable it

This seems easier to understand. Converting the various options of the env var into these can happen at the entrypoint of the entire application.

thebsdbox: That's fine.. I went down a convoluted route as we default blank to `metallb-system:config`. Will this default now live in the manifest?

deitch: No need. The default lives in `main.go` where it sets defaults on options.

This drags us back into the #117 discussion as to what the right way this should look to users is. But for now, the blank option is rendered in `main.go`, which deals with CLI flags and converts them into something computer-sane.

thebsdbox:
Closing this Proof of concept work.

PR #122, opened 2020-12-08
src IP annotation

deitch:
We already set annotations for peer ASN, local ASN, and peer IPs. This adds an annotation for the src IP, so you know which IP to use to source the peering request.

It also:

* standardizes the default annotation names
* fixes an error message to make it clearer


thebsdbox (approved):
Excellent, looking forward to leveraging this!

PR #123, opened 2020-12-08
add tagging for unique clusters

### LINKED ISSUE #120: EIP can conflict between 2 clusters in same project due to tag naming
ccm uses EM EIP tags to determine which service an EIP is attached to (or intended for). That way it can stop, start, etc. and the affiliation will be appropriate. You even can pre-assign an EIP (which @c0dyhi11 wants to do with Anthos).

The logic when syncing services inside ccm is roughly (ignoring removal, etc.):

1. Find all services of `type=LoadBalancer`
1. Filter out any that already have `Spec.LoadBalancerIP` set
1. Create two tags:
    * `usage=packet-ccm-auto`
    * `service=<service-identifier>` (see below)
1. Find any EIP that have the above tags set
    * if found, use it
    * if not, create one, add the tags
1. Assign the EIP by setting `Spec.LoadBalancerIP=<eip>`

The `<service-identifier>` is created as follows (pseudo-code):

```
identifier = sha256( serviceNamespace + "/" +serviceName )
```

We use sha256 instead of just the clear text, because some customers did not want the names of their services exposed on the project configuration; it should remain internal to the cluster only.

The problem arises when you have 2 or more clusters in the same project. If both have a `Service` with the same name, they will conflict. It will try to assign the EIP to both of them, which is not what you want.

deitch:
Fixes #120 ; also describes it in detail in the README.

cc @c0dyhi11

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ For each, CCM's behaviour is as follows:
 Environment variable overrides setting in kubernetes secret; the secret will be checked only if
 the environment variable is unset. Only if _both_ are unset will it default.
 
-If the `ConfigMap` is not disabled, then upon start, CCM does the following:
+If the `ConfigMap` is not disabled, then upon start, CCM does the following. EIP tag descriptions follow this list.
 
 1. Get the appropriate namespace and name of the `ConfigMap`, based on the rules above.
 1. If the `ConfigMap` does not exist, do nothing more.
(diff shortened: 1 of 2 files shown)


thebsdbox (approved):
`/lgtm`

README.md
```diff
@@ -237,22 +237,26 @@ If the `ConfigMap` is not disabled, then upon start, CCM does the following:
...
 modifies an existing `ConfigMap`. This can be deployed by the administrator separately, using the manifest
 provided in the releases page, or by any other manner.
 
-In all cases of tagging the IP address reservation, we tag the IP reservation with `usage="packet-ccm-auto"` and `service="<service-hash>"` where `<service-hash>` is the sha256 hash of `<namespace>.<service-name>`. We do this so that the name of the service does not leak out to Equinix Metal itself.
+In all cases of tagging the IP address reservation, we tag the IP reservation with the following tags:
 
-IP addresses always are created `/32`, and are tagged as `"packet-ccm-auto"`
+* `usage="packet-ccm-auto"`
+* `service="<service-hash>"` where `<service-hash>` is the sha256 hash of `<namespace>/<service-name>`. We do this so that the name of the service does not leak out to Equinix Metal itself.
+* `cluster=<clusterID>` where `<clusterID>` is the UID of the immutable `kube-system` namespace. We do this so that if someone runs two clusters in the same project, and there is one `Service` in each cluster with the same namespace and name, then the two EIPs will not conflict.
```
detiber, line 257: It looks like the other ccm implementations use a command line argument/config to make this configurable by the user (this is also something that we could template out pretty easily in cluster-api when deploying the CCM), this would prevent potential issues related to the UID changing if the cluster is restored from backup for any reason.

It looks like there is a shared configuration that is exposed through k8s.io/cloud-provider, which is what is used by the legacy ccm and the aws out-of-tree provider: https://github.com/kubernetes/cloud-provider/blob/master/config/v1alpha1/types.go#L60

It looks like vsphere went a different route and exposed their own flag: https://github.com/kubernetes/cloud-provider-vsphere/blob/22e85632bfe62f9a6afff8d2fc269f80eb9e7c78/cmd/vsphere-cloud-controller-manager/main.go

deitch: @detiber these are good reference points; thanks for bringing them here.

As I look through the [cloudprovider.Interface](https://godoc.org/k8s.io/cloud-provider#Interface), I see that the following all have methods that are passed a `clusterName string`:

* [Clusters](https://godoc.org/k8s.io/cloud-provider#Clusters)
* [LoadBalancer[(https://godoc.org/k8s.io/cloud-provider#LoadBalancer)
* [Routes](https://godoc.org/k8s.io/cloud-provider#Routes)

Are those the same [ClusterName](https://github.com/kubernetes/cloud-provider/blob/master/config/v1alpha1/types.go#L60) you linked to above, from the cloud-provider types? How is it set and used? It says [in the comment](https://github.com/kubernetes/cloud-provider-vsphere/blob/49f6445da72ec3c607125aa9aee89c8d830a8754/pkg/cloudprovider/vsphere/loadbalancer/lbprovider.go#L121):

```go
// Parameter 'clusterName' is the name of the cluster as presented to kube-controller-manager
```

Is that set via the `--cluster-name` parameter set on [kube-controller-manager](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/), which defaults to `"kubernetes"`?

Are you suggesting we just use that as the cluster name, and then let the creator decide on what name they want to use, and inherit it?

PR #124, opened 2020-12-08
switch the complex build and multi-arch to github actions

deitch:
Massively simplifies Makefile on the way, and adds Makefile help via `make help`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,9 @@ on:
     branches:
     - master
 
+env:
+  IMAGE_NAME: "packethost/packet-ccm"
+
 jobs:
   report:
     name: Report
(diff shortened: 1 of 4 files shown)


displague:
@deitch looks like a build failure:
```
failed to solve: rpc error: code = Unknown desc = failed to compute cache key: failed to walk /tmp/buildkit-mount235516694/dist/bin: lstat /tmp/buildkit-mount235516694/dist/bin: no such file or directory
Error: buildx call failed with: failed to solve: rpc error: code = Unknown desc = failed to compute cache key: failed to walk /tmp/buildkit-mount235516694/dist/bin: lstat /tmp/buildkit-mount235516694/dist/bin: no such file or directory
```

I like the move to the Docker GH actions, the diff seems sensible to me. I'll take a deeper look when we understand or correct the build failure.

deitch:
I think that is my error. I will dig into it (and rebase too).

deitch:
OK, hopefully that got it. The longer part was waiting for builds to run.

deitch:
Ah, CI is nice and happy now @displague

displague (approved):
This looks good to me. I think we'll find out if version stamping is working the automated way (🎲 ) and work from there.

I'm interested (non-blocking) if the `master` and `latest` images will share the same sha or if they will differ. If they differ I wonder if adding a build cache would help.

I think it also may be interesting to publish any origin branches on docker hub, so that we could test these CCMs.  This PR would be available as `packethost/packet-ccm:docker-action`, for example.

PR #125, opened 2020-12-09
universalize the LB control

### LINKED ISSUE #117: LoadBalancer enabling behaviours - what is the correct UX

deitch:
This is a PR for replacing the assumption that "everything is metallb" with a more universalized control of LBs.

The README describes well how this works.

Fixes #117


deitch:
So far, this only includes the README describing how the experience would be, as discussed in #117 and verbally with @thebsdbox

README.md
```diff
@@ -159,9 +159,20 @@ The Equinix Metal CCM works in one facility at a time. You can control which fac
...
-### BGP
+### Load Balancers
+
+Equinix Metal does not offer managed load balancers like [AWS ELB](https://aws.amazon.com/elasticloadbalancing/)
+or [GCP Load Balancers](https://cloud.google.com/load-balancing/). Instead, Equinix Metal CCM is able to interface
+with and configure external bare-metal loadbalancers.
+
+Depending on the type of loadbalancer configured, the CCM may set Kubernetes `Node` annotations, enable BGP on the project
+and specific devices, getting an Equinix Metal Elastic IP, set the `Spec.LoadBalancerIP` on the Kubernetes `Service`, and
+other activities. Each loadbalancer implementation may have unique ways of working.
```
displague, line 170: The text that follows in `#### BGP` starts us down a path of mental branching. `If the loadbalancer implementation requires`, `Each loadbalancer implementation`.

I wonder if we are moving into generic text too soon and we could settle with being specific for now:
`MetalLB requires BGP...` `Kube-VIP requires an EIP...`.

I also wonder if a markdown table (right here) might help. We _might_ be able to represent the supported loadbalancers, their various modes, capabilities, and requirements in table-format.

displague: We might also benefit from keeping the kube-vip and metallb docs separated, as unique markdown files.

deitch: I like the idea of separate markdown files quite a bit. I wasn't sure how to handle the bgp part, as it is a key part of design - CCM decides how to do bgp if needed - but specific to each. Hence the mental branching.

I will see if we can get some cleanup done on this, but also continue down this implementation path.

deitch:
Take two on this. It doesn't fully address it - it still does the BGP stuff whenever _any_ load-balancer is enabled - but it does universalize the control config, and makes the specific LB implementation pluggable.

I will address the doc question too shortly @displague, but this is a cleanup step forward, and should be mergeable as-is. We can pull out the BGP into optional in a separate PR.

With this in, we can plug in kube-vip as an option.

README.md
deitch: @displague I cleaned it up somewhat. I didn't yet break it out into separate files, since I haven't included kube-vip in this yet, but rather set up the ability to do it. Take another look please.

thebsdbox:
Awesome, going to take this for a test tomorrow. Will test on EM and at home with a BGP enabled router.

thebsdbox:
So I've taken it for a whirl and so far the `namespaces` `get` permission is missing. When trying to `expose` a service the reconcile is erroring as it's trying to do the controlplane EIP ?

deitch:
OK, will be sure to add `namespaces get` back in.

> When trying to expose a service the reconcile is erroring as it's trying to do the controlplane EIP ?

Show some log lines?

deitch:
Added the `namespaces get` permission.

thebsdbox:
EIP_control_plane errors
```
E1224 16:09:56.755832       1 cloud.go:268] failed to update and sync service for remove default/nginx: elastic ip tag is empty. Nothing to do
E1224 16:10:00.596734       1 cloud.go:260] failed to update and sync service for add default/nginx: elastic ip tag is empty. Nothing to do
E1224 16:10:11.531003       1 cloud.go:312] failed to update and sync services: elastic ip tag is empty. Nothing to do
```

I think if we add in either a `default://` or `bgp://`load-balancer option that has the loadbalancer interface we should be good.. Most of the functions won't need to to anything, but as long as the implementation skeleton exists then we should request EIPs from the API and update the service spec.

thebsdbox (commented):
Made a few suggestions around some small omissions (RBAC) etc..

The final piece of the puzzle from my (albeit one-sided view) is to add in the basic option for having the CCM just handle the API calls to the Equinix Metal IP address endpoints and the Kubernetes API (namely the services update).

deitch:
@thebsdbox I added the RBAC perm, and also an option for an `empty` LB config, which does nothing beyond the basic bgp enablement. That should be enough.

thebsdbox:
Works perfectly.. Perhaps implement `imagePullPolicy: Always` in the spec so I don't waste ages wondering why it isn't working (which was silly of me) :-D

deitch:
Added `imagePullPolicy` as requested @thebsdbox . Care to put a review on this?

thebsdbox (approved):
This is excellent, can't wait to start using the CCM 🥇

deitch:
And it is in.

@thebsdbox if this needs any additional controls or config to manage kube-vip, either open a PR for it, or if it needs discussion, open an issue and we will get it in together.

alice-sowerby:
Hey @deitch this seems like a big step forward - please could you summarise this and any other notable changes since [the last changelog](https://feedback.equinixmetal.com/changelog/kubernetes-ccm-improvements) so that I can release another changelog announcement?

PR #126, opened 2020-12-15
Makefile: add targets for building multi-arch Docker manifests

### LINKED ISSUE #104: Makefile: platforms joining doesn't work
Currently, trying to run locally `make release` fails with the following error for me:
```
bgp-node-selector-amd64: digest: sha256:bf058f9eef5ce126feb1d8d097f2d6597917d997481c770fd132772125e37212 size: 739
make[2]: Leaving directory '/home/invidian/repos/kinvolk/packet-ccm'
# path to credentials based on manifest-tool's requirements here https://github.com/estesp/manifest-tool#sample-usage
/home/invidian/go/bin/manifest-tool push from-args --platforms linux/arm64 linux/amd64, --template quay.io/invidian/packet-ccm:bgp-node-selector-ARCH --target quay.io/invidian/packet-ccm:bgp-node-selector
FATA[0000] You must specify all three arguments --platforms, --template and --target
make[1]: *** [Makefile:198: push-manifest] Error 1
make[1]: Leaving directory '/home/invidian/repos/kinvolk/packet-ccm'
make: *** [Makefile:267: release] Error 2
```

invidian:
So multi-arch manifests can be also built locally, not only on CI.

One can now run:
```
make image-manifest-all-push BUILD_IMAGE=quay.io/foo/packet-ccm BUILD_TAG=bar
```

To build and push all required images, build the manifest and push it as well.

Closes #104.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,8 @@ VERSION := $(RELEASE_TAG)-$(VERSION)
 endif
 GO_FILES := $(shell find . -type f -not -path './vendor/*' -name '*.go')
 BUILD_TAG ?= latest
+TAGGED_IMAGE ?= $(BUILD_IMAGE):$(BUILD_TAG)
+TAGGED_ARCH_IMAGE ?= $(TAGGED_IMAGE)-$(ARCH)
 LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)"'
 
 # which arches can we support
@@ -161,8 +163,8 @@ sub-image-%:
 	@$(MAKE) ARCH=$* image
 
 image: ## make the image for a single ARCH
-	docker buildx build --load -t $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH) -f Dockerfile --platform $(OS)/$(ARCH) .
-	echo "Done. image is at $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH)"
+	docker buildx build --load -t $(TAGGED_ARCH_IMAGE) -f Dockerfile --platform $(OS)/$(ARCH) .
+	echo "Done. image is at $(TAGGED_ARCH_IMAGE)"
 
 # Targets used when cross building.
 .PHONY: register
--- a/Makefile
+++ b/Makefile
@@ -166,6 +166,26 @@ image: ## make the image for a single ARCH
 	docker buildx build --load -t $(TAGGED_ARCH_IMAGE) -f Dockerfile --platform $(OS)/$(ARCH) .
 	echo "Done. image is at $(TAGGED_ARCH_IMAGE)"
 
+push-all: $(addprefix push-arch-, $(ARCHES)) ## Push all built images.
+push-arch-%:
+	@$(MAKE) ARCH=$* push
+
+push: image ## Push image to registry for a single ARCH.
+	docker push $(TAGGED_ARCH_IMAGE)
+
+manifest-push: manifest-all ## Make single image manifest for all supported ARCH and push it to registry.
+	docker manifest push $(TAGGED_IMAGE)
+
+manifest-all: manifest-create $(addprefix manifest-annotate-arch-, $(ARCHES)) ## Annotate docker manifest with all supported ARCH.
+manifest-annotate-arch-%:
+	@$(MAKE) ARCH=$* manifest-annotate
+
+manifest-annotate:
+	docker manifest annotate $(TAGGED_IMAGE) $(TAGGED_ARCH_IMAGE) --arch=$(ARCH) --os=$(OS)
+
+manifest-create: push-all ## Creates Docker manifest for all created images.
+	docker manifest create $(TAGGED_IMAGE) $(addprefix --amend $(TAGGED_IMAGE)-, $(ARCHES))
+
 # Targets used when cross building.
 .PHONY: register
 # Enable binfmt adding support for miscellaneous binary formats.
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ ifneq (,$(RELEASE_TAG))
 VERSION := $(RELEASE_TAG)-$(VERSION)
 endif
 GO_FILES := $(shell find . -type f -not -path './vendor/*' -name '*.go')
-FROMTAG ?= latest
+BUILD_TAG ?= latest
 LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)"'
 
 # which arches can we support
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -161,8 +161,31 @@ sub-image-%:
 	@$(MAKE) ARCH=$* image
 
 image: ## make the image for a single ARCH
-	docker buildx build --load -t $(BUILD_IMAGE):latest-$(ARCH) -f Dockerfile --platform $(OS)/$(ARCH) .
-	echo "Done. image is at $(BUILD_IMAGE):latest-$(ARCH)"
+	docker buildx build --load -t $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH) -f Dockerfile --platform $(OS)/$(ARCH) .
```
deitch, line 164: This phrase `$(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH)` is repeated in several places (4 at last count) and the one without `-$(ARCH)` in several more (also 4, I think). These should be replaced with a single var:

```makefile
TAGGED_IMAGE ?= $(BUILD_IMAGE):$(BUILD_TAG)
TAGGED_ARCH_IMAGE ?= $(TAGGED_IMAGE)-$(ARCH)
```

and then use those

Makefile
```diff
@@ -161,8 +161,31 @@ sub-image-%:
...
+	echo "Done. image is at $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH)"
+
+image-all-push: $(addprefix sub-image-push-, $(ARCHES)) ## Push all built images.
+sub-image-push-%:
+	@$(MAKE) ARCH=$* image-push
+
+image-push: ## Push image to registry for a single ARCH.
```
deitch, line 171: 1. Is there any reason not to call this just `push`? The names of the targets are getting mighty long and confusing.
2. Shouldn't this have `image` as a dependency?

Makefile
```diff
@@ -161,8 +161,31 @@ sub-image-%:
 	@$(MAKE) ARCH=$* image
 
 image: ## make the image for a single ARCH
-	docker buildx build --load -t $(BUILD_IMAGE):latest-$(ARCH) -f Dockerfile --platform $(OS)/$(ARCH) .
-	echo "Done. image is at $(BUILD_IMAGE):latest-$(ARCH)"
+	docker buildx build --load -t $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH) -f Dockerfile --platform $(OS)/$(ARCH) .
+	echo "Done. image is at $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH)"
+
+image-all-push: $(addprefix sub-image-push-, $(ARCHES)) ## Push all built images.
```
deitch, line 167: If the other becomes `push`, then this can be `push-all` and the next one `sub-push-%:`

Makefile
```diff
@@ -161,8 +161,31 @@ sub-image-%:
...
+sub-image-push-%:
+	@$(MAKE) ARCH=$* image-push
+
+image-push: ## Push image to registry for a single ARCH.
+	docker push $(BUILD_IMAGE):$(BUILD_TAG)-$(ARCH)
+
+image-manifest-all: image-manifest-create image-manifest-annotate-all ## Make single image manifest for all supported ARCH.
```
deitch, line 174: Why is there `manifest-all` and `manifest-create`? Since `-all` is for "all arches", isn't `manifest` by definition for "all"?

deitch: I think you can simplify the multiple `manifest` targets into fewer. Just have a single one that creates a manifest for all know arches.

deitch (changes requested):
This is fine overall. It needs some simplification and cleanup, especially in terms of naming the targets and the number of them.

Makefile
invidian: >Is there any reason not to call this just push? The names of the targets are getting mighty long and confusing.

Well, it puts all Docker image related targets into a `image-` prefix, which I think gives nice grouping. perhaps `sub-image-push` could be changed to `image-sub-push` or `image-push-sub` ? Or Maybe `image-push-arch` ?

>Shouldn't this have image as a dependency?

Yes :+1:

Makefile
invidian: >Why is there manifest-all and manifest-create? Since -all is for "all arches", isn't manifest by definition for "all"?

This is because order of operation is important here. It must be:
1. Ensure all images are pushed.
2. Create the manifest.
3. Annotate it for all architectures.

I tried to simplify it, please have a look :)

Makefile
deitch: > it puts all Docker image related targets into a image- prefix, which I think gives nice grouping. perhaps sub-image-push could be changed to image-sub-push or image-push-sub ? Or Maybe image-push-arch ?

I get it, but I don't think it is necessary. It buys us unnecessary flexibility at the price of additional complexity. What else are we going to `push` from here? What else are we going to do with `manifest`?

Makefile
deitch: I actually never have used `docker manifest create` and `docker manifest annotate`; I have been using manifest-tool since day 1: just push a bunch of images, and then create an index and push it in a single command.

Unfortunately, the documentation for `docker manifest create` and `docker manifest annotate` are sparse (to put it mildly). 

I would have thought `manifest create` makes it locally, and `manifest annotate adds data to the specific entry in the manifests list in the index, and `manifest push` pushes it. But if I try to do `manifest create`, I get registry related errors?

Either way, is there any reason for all of the targets? Why not just `make manifest` and have it create, annotate and push? If someone wants to do something manually other than that, they can.

Makefile
invidian: `manifest` might be related to Kubernetes manifests, which is more likely than Docker manifest I think. But yeah, I don't insist on that, let me change it.

Makefile
invidian: >I would have thought manifest create makes it locally, and manifest annotate adds data to the specific entry in the manifests list in the index, and manifest pushpushes it. But if I try to domanifest create`, I get registry related errors?

Yeah, this is how it works. However, when creating the manifest, I think it takes the image information from the registry, as manifest only contains a references to other images, which must be already pushed.

>Either way, is there any reason for all of the targets? Why not just make manifest and have it create, annotate and push? If someone wants to do something manually other than that, they can.

So `annotate` is a separate target so we can use `addprefix` expansion pattern which is used for building the image. But creating the manifest must create before, hence separate target to run it.

And I think `push` target can be merged with `all`, as running it will push individual images anyway.

If you could help me with simplifying it, that would be great, as I'm not sure how passing ARCH works exactly etc.

deitch: Eh, I just tried and could not come up with a cleaner way. Minutely, but not seriously. I dislike having it in the Makefile _and_ in the GitHub Actions. But so be it.

Let's remove the `image-*` prefixes to make it easier to understand and take it as is.

invidian: >Eh, I just tried and could not come up with a cleaner way. Minutely, but not seriously. I dislike having it in the Makefile and in the GitHub Actions. But so be it.

This is why I dislike GitHub Actions a bit. IMO CI should only automate tasks which can be executed locally. So if your CI breaks, you can still perform all maintenance tasks.

I removed the `image-*` prefixes and merged commits. I'm testing now and if everything works fine, will push and this will be ready to merge.

Thanks again for looking into that!

displague: @invidian Have you tried [`act`](https://github.com/nektos/act)?

I run GH Actions locally with: `act -P ubuntu-latest=nektos/act-environments-ubuntu:18.04 -b` 

I've run into some Actions that don't work as expected, most recently with local mount points from the Mac needing to get passed into actions that invoke `docker`.

displague:
This looks like it was cleaned up as requested, @deitch. I defer to your judgement on this PR.

invidian:
BTW, I used this feature when working on Gardener issues and it was super helpful to be able to quickly build a testing image for myself and have it uploaded automagically :smile:

deitch (approved):

deitch:
This needs to be rebased, and then I will merge it in.

invidian:
Rebased.

deitch (approved):

invidian:
Thanks for merging!

PR #128, opened 2020-12-31
Fix boolean logic in auth token debug info

johananl:
We want to print `<masked>` when the auth token string is NOT empty.

cc: @deitch


deitch (approved):
Oops. 😁

PR #138, opened 2021-01-27
fix tagging in release

deitch:


gianarb (approved):
You rock!! :D Take a break, make a 🍵 !

PR #144, opened 2021-01-29
Switch to klog v2

### LINKED ISSUE #143: Unable to increase CCM log verbosity
It seems adding `-v=2` to CCM deployment, even though accepted, has no effect and does not increase the verbosity of the process. This effectively blocks more thorough debugging and makes extensive logging in the code base useless.

I suspect this might be because of misconfigured `klog`, as e.g. I don't see calls to `klog.InitFlags(nil)` mentioned in https://github.com/kubernetes/klog#how-to-use-klog.

invidian:
To fix switching log verbosity, as right now adding -v=10 has no effect.

Closes #143

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,7 +15,7 @@ import (
 	"k8s.io/component-base/logs"
 	_ "k8s.io/component-base/metrics/prometheus/clientgo" // for client metric registration
 	_ "k8s.io/component-base/metrics/prometheus/version"  // for version metric registration
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/cmd/cloud-controller-manager/app"
 
 	"github.com/equinix/cloud-provider-equinix-metal/metal"
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	k8s.io/client-go v0.19.4
 	k8s.io/cloud-provider v0.19.4
 	k8s.io/component-base v0.19.4
-	k8s.io/klog v1.0.0
+	k8s.io/klog/v2 v2.5.0
 	k8s.io/kubernetes v1.19.4
 )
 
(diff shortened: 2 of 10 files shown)


main.go
```diff
@@ -15,7 +15,7 @@ import (
 	"k8s.io/component-base/logs"
 	_ "k8s.io/component-base/metrics/prometheus/clientgo" // for client metric registration
 	_ "k8s.io/component-base/metrics/prometheus/version"  // for version metric registration
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
```
displague, line 18: Supporting docs on why this does not need to be aliased is at https://blog.golang.org/using-go-modules:

> Each different major version (v1, v2, and so on) of a Go module uses a different module path: starting at v2, the path must end in the major version. In the example, v3 of rsc.io/quote is no longer rsc.io/quote: instead, it is identified by the module path rsc.io/quote/v3. This convention is called [semantic import versioning](https://research.swtch.com/vgo-import), and it gives incompatible packages (those with different major versions) different names. In contrast, v1.6.0 of rsc.io/quote should be backwards-compatible with v1.5.2, so it reuses the name rsc.io/quote. (In the previous section, rsc.io/sampler v1.99.99 should have been backwards-compatible with rsc.io/sampler v1.3.0, but bugs or incorrect client assumptions about module behavior can both happen.)

displague (dismissed):
Supporting docs on why `k8s.io/klog/v2` does not need to be aliased to `klog` is at https://blog.golang.org/using-go-modules:

> Each different major version (v1, v2, and so on) of a Go module uses a different module path: starting at v2, the path must end in the major version. In the example, v3 of rsc.io/quote is no longer rsc.io/quote: instead, it is identified by the module path rsc.io/quote/v3. This convention is called [semantic import versioning](https://research.swtch.com/vgo-import), and it gives incompatible packages (those with different major versions) different names. In contrast, v1.6.0 of rsc.io/quote should be backwards-compatible with v1.5.2, so it reuses the name rsc.io/quote. (In the previous section, rsc.io/sampler v1.99.99 should have been backwards-compatible with rsc.io/sampler v1.3.0, but bugs or incorrect client assumptions about module behavior can both happen.)

invidian:
Actually, updated to latest version of `klog/v2`.

deitch (approved):

PR #153, opened 2021-02-15
handle metallb configmap without config

### LINKED ISSUE #151: Metallb default ConfigMap has no data.config field
By default the `configMap` applied by the metallb installer doesn't create the `config` key in the `data`. The CCM will report that no config exists with the name `config`... Ideally if the CCM finds the `configMap` it should be able to create the `config` KV before populating it.

```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: "I'm missing"
```

deitch:
Fixes #151

### DIFF
--- a/metal/loadbalancers/metallb/metallb.go
+++ b/metal/loadbalancers/metallb/metallb.go
@@ -3,7 +3,6 @@ package metallb
 import (
 	"context"
 	"encoding/json"
-	"errors"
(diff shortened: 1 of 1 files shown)


thebsdbox (approved):
Wow *this* was quick.

thebsdbox (approved):

PR #158, opened 2021-04-14
Add backward compatibility for providerID convention

detiber:
Allows matching devices based on the old providerID convention (packet://).

This is to help enable migrating Cluster API-based clusters to newer versions of the CCM.

### DIFF
--- a/metal/cloud.go
+++ b/metal/cloud.go
@@ -19,6 +19,11 @@ import (
 
 const (
 	providerName string = "equinixmetal"
+
+	// deprecatedProviderName is used to provide backward compatibility support
+	// with previous versions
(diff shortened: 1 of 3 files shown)


deitch:
I like this in principle. Smart move. But CI checks are failing. Can you see why?

detiber:
> I like this in principle. Smart move. But CI checks are failing. Can you see why?

It was catching the fact that I mistakenly used `||` instead of `&&` in the condition, should be fixed now.

I'm glad I added the test cases, because it caught my silly mistake from creating this PR while multi-tasking :)

deitch (approved):
as soon as CI is green, we can merge in

PR #159, opened 2021-04-15
fix handling of metallb custom configmap and document it

deitch:
There were 2 issues:

1. The README didn't make it clear that when providing config for a load balancer, you need to provide everything in path, not host, so, e.g. `metallb:///namespace/configname` and not `metallb://namespace/configname` (three slashes instead of two); fixes the doc
2. Fixes parsing of the above config information, and changes it to use `/` as a separator; `:` causes too many issues, and is un-URL-ish

### DIFF
--- a/metal/loadbalancers/metallb/metallb.go
+++ b/metal/loadbalancers/metallb/metallb.go
@@ -28,7 +28,14 @@ type LB struct {
 
 func NewLB(k8sclient kubernetes.Interface, config string) *LB {
 	var configmapnamespace, configmapname string
-	cmparts := strings.SplitN(config, ":", 2)
+	// it may have an extra slash at the beginning or end, so get rid of it
+	if strings.HasPrefix(config, "/") {
+		config = config[1:]
+	}
+	if strings.HasSuffix(config, "/") {
+		config = config[:len(config)-1]
+	}
+	cmparts := strings.SplitN(config, "/", 2)
 	if len(cmparts) >= 2 {
 		configmapnamespace, configmapname = cmparts[0], cmparts[1]
 	}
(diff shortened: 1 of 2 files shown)


detiber (commented):
The changes look fine to me, but are there any concerns around upgrades?

deitch:
There would be, but the old pattern actually didn't work. 😃

detiber:
> There would be, but the old pattern actually didn't work.

Good enough for me :)

detiber (approved):

metal/loadbalancers/metallb/metallb.go
```diff
@@ -28,7 +28,14 @@ type LB struct {
...
-	cmparts := strings.SplitN(config, ":", 2)
+	// it may have an extra slash at the beginning or end, so get rid of it
+	if strings.HasPrefix(config, "/") {
+		config = config[1:]
+	}
+	if strings.HasSuffix(config, "/") {
+		config = config[:len(config)-1]
+	}
+	cmparts := strings.SplitN(config, "/", 2)
 	if len(cmparts) >= 2 {
```
displague, line 39: nitpic: `>` doesn't really apply anymore

displague (approved):

displague:
> There would be, but the old pattern actually didn't work. 😃

otoh - if anyone was scratching their head about why this wasn't working as previously documented, it would just start working on their next image update.

I'm fine with the `///` for what it's worth.

metal/loadbalancers/metallb/metallb.go
deitch: Oh that is a good point! It doesn't break anything, and I don't want to hold this up when under real-world usage issues, so I am going to merge this in (as you already approved) and will get that in the next changes.

PR #163, opened 2021-04-18
set md5 passphrase base64-encoded in annotations, add to metallb configmap

deitch:
I am not done with testing this yet, but it is ready for eyes-on review.

This places the md5 password string, however it appears when getting device info from the Equinix Metal API, in a few places:

* an annotation, base64-encoded. This doesn't provide real security, but it is better than being able to see it at a glance.
* in the metallb configmap in the right place for the peer

It also adds supportive details:  an env var for configuring the annotation name, a default annotation name, documentation of the above.

It also cleans up the README a bit: removing a duplicate entry in the configuration table, and clarifies that the env var for local ASN and BGP password are solely for requesting it when enabling on the project; when actually setting it on nodes, it uses the ones retrieved from the EM API. This behaviour is unchanged; just docs clarification of the above.

cc @enkelprifti98

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,17 +148,23 @@ This section lists each configuration option, and whether it can be set by each 
 | Facility |    | `METAL_FACILITY_NAME` | `facility` | read metadata on host on which CCM is running, else error |
 | Base URL to Equinix API |    |    | `base-url` | Official Equinix Metal API |
 | Load balancer setting |   | `METAL_LB` | `loadbalancer` | none |
-| BGP ASN for each cluster node |   | `METAL_LOCAL_ASN` | `localASN` | `65000` |
-| BGP ASN for upstream peer |   | `METAL_PEER_ASN` | `peerASN` | `65530` |
+| BGP ASN for cluster nodes when enabling BGP on the project |   | `METAL_LOCAL_ASN` | `localASN` | `65000` |
+| BGP passphrase to use when enabling BGP on the project |   | `METAL_BGP_PASS` | `bgpPass` | `""` |
 | Kubernetes annotation to set node's BGP ASN |   | `METAL_ANNOTATION_LOCAL_ASN` | `annotationLocalASN` | `"metal.equinix.com/node-asn"` |
 | Kubernetes annotation to set BGP peer's ASN |   | `METAL_ANNOTATION_PEER_ASNS` | `annotationPeerASNs` | `"metal.equinix.com/peer-asn"` |
 | Kubernetes annotation to set BGP peer's IPs |   | `METAL_ANNOTATION_PEER_IPS` | `annotationPeerIPs` | `"metal.equinix.com/peer-ip"` |
 | Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
-| Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
+| Kubernetes annotation to set BGP MD5 password, base64-encoded (see security warning below) |   | `METAL_ANNOTATION_BGP_PASS` | `annotationBGPPass` | `"metal.equinix.com/bgp-pass"` |
 | Tag for control plane Elastic IP |    | `METAL_EIP_TAG` | `eipTag` | No control plane Elastic IP |
 | Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | `6443` |
-| Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | `6443` |
 | Filter for cluster nodes on which to enable BGP |    | `METAL_BGP_NODE_SELECTOR` | `bgpNodeSelector` | All nodes |
+
+<u>Security Warning</u>
+Including your project's BGP password, even base64-encoded, may have security implications. Because Equinix Metal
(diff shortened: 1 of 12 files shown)


metal/constants.go
```diff
@@ -10,6 +10,7 @@ const (
 	DefaultAnnotationPeerASNs = "metal.equinix.com/peer-asn"
 	DefaultAnnotationPeerIPs  = "metal.equinix.com/peer-ip"
 	DefaultAnnotationSrcIP    = "metal.equinix.com/src-ip"
+	DefaultAnnotationBGPPass  = "metal.equinix.com/bgp-pass"
```
displague, line 13: maybe we should include `-insecure` in this name

displague:
While looking at this, I had to wonder why we need to make the BGP password part of an annotation.
What is reading the node annotation and why is this the best way for that application to access that information. 

Did the user have the opportunity to configure that app with this password in another way? (helm chart, configmap, etc).

Is the app that needs the BGP password outside of the the CCMs control?
If the app is in the CCMs control (config management), can the CCM provide this information in a more direct manneter?

displague:
The list of behaviors in https://github.com/equinix/cloud-provider-equinix-metal#kube-vip (and the other LB options) should include BGP password with the other BGP settings that are handled. (bullet "2")

deitch:
> While looking at this, I had to wonder why we need to make the BGP password part of an annotation.
> What is reading the node annotation and why is this the best way for that application to access that information.
>
> Did the user have the opportunity to configure that app with this password in another way? (helm chart, configmap, etc).
>
> Is the app that needs the BGP password outside of the the CCMs control?
> If the app is in the CCMs control (config management), can the CCM provide this information in a more direct manneter?

It is easier to answer this all at once. The flow is as follows:

1. (optional) User enables bgp, with password, outside of CCM
2. CCM enables bgp, if it is not enabled
3. for each node it encounters:
    1. CCM reads the BGP config (src IP, peer IPs, local ASN, peer ASN, BGP pass) from the EM API
    2. CCM places that config  on annotations
4. Consuming services (e.g. metallb, kube-vip, etc.) read that config from annotations on the node on which it is running and apply it

The annotation is the "message bus" between CCM and kube-vip, which is out of its control. The user doesn't set the BGP password that is passed to the nodes when running CCM; CCM reads it from the EM API for each device.

That config is pretty much unique per node. In practice, there often is lots of overlap (ASNs, upstream IPs sometimes, etc.), but there need not be.

Granted, there could be "other channels" for passing that information along - metallb uses a configmap, although there is a long-open PR to replace (or at least augment, with an eye towards replacing in the future) that in favour of annotations - but those "other channels" have a lot of shortcomings:

* they are specific to each type (metallb's unique configmap, etc.) that CCM would then need to know and keep up to date with each one. This has been painful in the past, and lots of duplication.
* they sometimes don't scale well
* they can lead to conflicts (we have had cases of metallb controller and CCM both trying to modify the configmap at the same time; ouch).

My ideal, as I envision it now, is that there is a per-node Secret or ConfigMap, part of core k8s API, which contains additional information of any type, with some pre-defined standard keys. Everyone knows to use them, and they do.

Since we don't have that now, annotations are the cleanest way we can think of.

thebsdbox (approved):
`/lgtm` 

I guess we'll need to track this one as its something that we will need a better resolution for in the future.

deitch:
@thebsdbox thanks.

I want to hold on merging this until @displague has had a chance to comment on the answers to his questions. I also want to do one more run of testing on it. (and wouldn't mind feedback from @gianarb and @detiber )

README.md
```diff
@@ -148,16 +148,15 @@ This section lists each configuration option, and whether it can be set by each
...
-| BGP ASN for upstream peer |   | `METAL_PEER_ASN` | `peerASN` | `65530` |
+| BGP ASN for cluster nodes when enabling BGP on the project |   | `METAL_LOCAL_ASN` | `localASN` | `65000` |
+| BGP passphrase to use when enabling BGP on the project |   | `METAL_BGP_PASS` | `bgpPass` | `""` |
 | Kubernetes annotation to set node's BGP ASN |   | `METAL_ANNOTATION_LOCAL_ASN` | `annotationLocalASN` | `"metal.equinix.com/node-asn"` |
 | Kubernetes annotation to set BGP peer's ASN |   | `METAL_ANNOTATION_PEER_ASNS` | `annotationPeerASNs` | `"metal.equinix.com/peer-asn"` |
 | Kubernetes annotation to set BGP peer's IPs |   | `METAL_ANNOTATION_PEER_IPS` | `annotationPeerIPs` | `"metal.equinix.com/peer-ip"` |
 | Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
-| Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
+| Kubernetes annotation to set BGP MD5 password, base64-encoded |   | `METAL_ANNOTATION_BGP_PASS` | `annotationBGPPass` | `"metal.equinix.com/bgp-pass"` |
```
detiber, line 157: Can we add a footnote here, or otherwise add a warning to this doc around exposing this via an annotation on the node to pre-preemptively address concerns folks may have?

deitch: Definitely a good call. Adding.

deitch: Take a look now, @detiber

detiber: lgtm, thanks!

thebsdbox (dismissed):
I approve, especially with the security warning in the readme.md

deitch:
Found one missing area where it needed to pass the password to metallb. Fixed.

thebsdbox (approved):
👍

PR #166, opened 2021-05-13
Add Helm chart

### LINKED ISSUE #92: Create and manage Helm chart for packet-ccm
It would be nice if there is a helm chart for the deployment and loadbalancer converted to approproate files. What do you think @deitch?

clrxbl:
Figured there's probably some interest in a Helm chart. Open for feedback as this is the first public chart I've created.
Github Actions needs fixing & the chart itself needs to be tested on an actual Equinix Metal Kubernetes cluster, will be doing so later.

Fixes #92


displague:
This looks great @clrxbl! What changes do you expect to make?
We'll need to publish the chart with the releases (in the release artifacts), I assume, and add notes about consuming this to the docs.

Questions for future PRs:
* Should the static yaml files in the repo be generated using helm?
* Use this action? https://github.com/helm/chart-testing-action

clrxbl:
For consistency purposes it would make sense to generate the static deployment manifests with helm template & distribute those for users that don't want to use Helm.

The Github Actions CI workflows would definitely need some changes, I'm not sure if I should be touching those within this PR though? Not sure if someone else wants/plans on doing that.

Documentation in general is something that needs touching up on (I might open another PR after this one to improve on that) as it took me some time to figure out MetalLB & control plane load balancing (still having issues with that) due to missing documentation.

clrxbl:
I've addd some changes to require certain config options & automatically roll the deployment on secret changes.

deitch:
I am excited to have this @clrxbl ; thank you for jumping in. 

> some interest

More than "some"!

deitch:
> Should the static yaml files in the repo be generated using helm?

100% yes. Not in this PR; one PR to get it in, another to consolidate. But yes.

> We'll need to publish the chart with the releases (in the release artifacts)

Same. We can do it in the future "refactor deployments" PR, or perhaps in a separate one.

deitch:
@clrxbl let us know when you are ready for reviews. No real need for CI on this, but we still will stick with "always run CI"

clrxbl:
I think the Helm chart is ready for use so you can start reviewing it.

deitch (changes requested):
I didn't compare the resources - Deployment, ClusterRole, ClusterRoleBinding, ServiceAccount, Service - to those of the original manifests in `deploy/`, so I assume these are matches.

I like the enable/disable control, and the check for required elements in the secret.

A few things stand out to me.

First, can we document how to use the helm chart? The README only references the deployment, but we should describe how to use helm to deploy as well.

Second, appVersion is hard-coded into here as `3.1.0`. The static manifests, on the other hand, do not have it hard-coded in. Instead they have [RELEASE_TAG](https://github.com/equinix/cloud-provider-equinix-metal/blob/5de7790e19e94d9608a42aea67631bdf6c1955b1/deploy/template/deployment.yaml#L36) which  CI replaces [here](https://github.com/equinix/cloud-provider-equinix-metal/blob/5de7790e19e94d9608a42aea67631bdf6c1955b1/.github/workflows/release.yaml#L53) for the current tag, this creating the correct deployment artifacts.

I know I said earlier that we should hold on CI, but I realized that this means the version is hard-coded in here. We either should update it here (can be using `sed` as it does in the workflow, or helm tools if easier), or we should follow up immediately with a separate PR. We finally got past manually updating the release tag in `deploy/template/deployment.yaml` (ask how many times we forgot, or mismatched with the git tag; better yet, please do not), we wouldn't want to reintroduce it.

clrxbl:
I've modified the appVersion to default to "master".

I've modified the release workflow to use https://github.com/helm/chart-releaser-action which seems promising. I've also made it run a sed command just like in the non-Helm release workflow.

deitch:
> I've modified the release workflow to use https://github.com/helm/chart-releaser-action which seems promising. 

It is, however, somewhat aggressive. With each "merge to master", it checks if the helm chart has changed, and if it has, actually creates a new release on GitHub. I don't think we want to turn that level of control over to it.

Also, [this line](https://github.com/equinix/cloud-provider-equinix-metal/pull/166/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR100) will change the file in the repository on the fly (other than the issue that it points to `deploy/helm/Chart.yaml` instead of `deploy/chart/Chart.yaml`), which will lead to a dirty git repo and possible knock-down effects.

I am going to backtrack on myself. Let's get rid of the pipeline change, leave the `appVersion: master` in there, and let's get this merged in. We can deal with pipeline automation as a separate step.

clrxbl:
> > I've modified the release workflow to use https://github.com/helm/chart-releaser-action which seems promising.
> 
> It is, however, somewhat aggressive. With each "merge to master", it checks if the helm chart has changed, and if it has, actually creates a new release on GitHub. I don't think we want to turn that level of control over to it.
> 
> Also, [this line](https://github.com/equinix/cloud-provider-equinix-metal/pull/166/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaR100) will change the file in the repository on the fly (other than the issue that it points to `deploy/helm/Chart.yaml` instead of `deploy/chart/Chart.yaml`), which will lead to a dirty git repo and possible knock-down effects.
> 
> I am going to backtrack on myself. Let's get rid of the pipeline change, leave the `appVersion: master` in there, and let's get this merged in. We can deal with pipeline automation as a separate step.

I've reverted the workflow changes.

deitch (approved):

deitch:
As soon as CI is green, we can merge this in, then open a tracking issue for releases.

PR #167, opened 2021-05-13
add docs about BGP peers

### LINKED ISSUE #165: include required static routes in the documentation or a daemonset
Clusters choosing to use MetalLB must add static routes:

```sh
GATEWAY_IP=$(curl https://metadata.platformequinix.com/metadata | jq -r ".network.addresses[] | select(.public == false) | .gateway")
ip route add 169.254.255.1 via $GATEWAY_IP
ip route add 169.254.255.2 via $GATEWAY_IP
```

This is not covered in the documentation.  It should be covered especially for those users that want to bring their own MetalLB of KubeVIP and will not be using the documented manifests.

Questions:
* Is this necessary for KubeVIP too?
* Could this be bundled as a DaemonSet to simplify the experience? If so, should this be added to the manifest-guided install, guarded behind flags?
* Is a DaemonSet run too late in the process? Do these routes need to be defined above Kubernetes, and if so, should a Kubernetes manifest be used to set that local configuration?

deitch:
Fixes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,57 @@ Recommended versions of Equinix Metal CCM based on your Kubernetes version:
 * Equinix Metal CCM version v0.0.4 supports Kubernetes version >=v1.10
 * Equinix Metal CCM version v1.0.0+ supports Kubernetes version >=1.15.0
 
+### BGP
+
+If you plan on using a BGP, for example for a BGP-based load balancer, you _may_ need to set static routes on your hosts.
+
+Details about BGP can be found in [the official Equinix Metal BGP documentation](https://metal.equinix.com/developers/docs/networking/local-global-bgp/#server-host-configuration).
+
+Equinix Metal facilities provide BGP peers at certain addresses, normally `169.254.255.1` and `169.254.255.2`. These are available
+in the host configuration via the Equinix Metal API, as well as the [metadata](https://metal.equinix.com/developers/docs/servers/metadata/)
+on each host.
+
+In order for BGP peering to work, the upstream BGP peers _must_ receive the packets from your device's _private_ IP address.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,6 +24,53 @@ Recommended versions of Equinix Metal CCM based on your Kubernetes version:
...
+* your BGP peer addresses
+
+A sample method:
+
+```bash
+GATEWAY_IP=$(curl https://metadata.platformequinix.com/metadata | jq -r '.network.addresses[] | select(.public == false and .address_family == 4) | .gateway')
+PEERS=$(curl https://metadata.platformequinix.com/metadata | jq -r '.bgp_neighbors[0].peer_ips[]')
+for i in ${PEERS};do
+ip route add ${i} via $GATEWAY_IP
+ip route add ${i} via $GATEWAY_IP
```
clrxbl, line 70: Should this be a duplicate?

deitch: Good catch. Fixing now.

README.md
```diff
@@ -24,6 +24,57 @@ Recommended versions of Equinix Metal CCM based on your Kubernetes version:
...
+
+* use BGP software that knows how to set the source address on a packet
+* set static routes on your host
+
+#### BGP Software
+
+Some implementations of BGP software support setting a source address for BGP peering packets, including
+[bird](https://bird.network.cz) and [kube-vip](https://kube-vip.io).
+
+CCM helps in this regard. It reads the information about the peers and the correct source address for the device from
```
displague, line 50: Do we want to start calling ourselves "CPEM" in the README? I assume we have a leading "Cloud Provider Equinix Metal (CPEM)" somewhere to allow us to use the short version throughout.

displague (approved):

PR #168, opened 2021-05-14
handle error safely in healthcheck

deitch:
Due to how the healthcheck logic worked, it tried to do `rest.Body.Close()` even if `resp.Body` was `nil`. This ensures we only close if `err == nil && `resp.Body != nil`.

Thanks to @displague for figuring out the logic problem.

### DIFF
--- a/metal/eip_controlplane_reconciliation.go
+++ b/metal/eip_controlplane_reconciliation.go
@@ -106,9 +106,13 @@ func (m *controlPlaneEndpointManager) reconcileNodes(ctx context.Context, nodes 
 		return err
 	}
 	resp, err := m.httpClient.Do(req)
+	// if there was no error, ensure we close
+	if err == nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
 	if err != nil || resp.StatusCode != http.StatusOK {
 		if err != nil {
-			klog.Errorf("http client error during healthcheck. err \"%s\"", err)
+			klog.Errorf("http client error during healthcheck, will try to reassign to a healthy node. err \"%s\"", err)
 		}
 		// filter down to only those nodes that are tagged as control plane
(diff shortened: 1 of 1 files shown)


metal/eip_controlplane_reconciliation.go
```diff
@@ -106,9 +106,13 @@ func (m *controlPlaneEndpointManager) reconcileNodes(ctx context.Context, nodes
 		return err
 	}
 	resp, err := m.httpClient.Do(req)
+	// if there was no error, ensure we close
+	if err == nil && resp.Body != nil {
```
displague, line 110: sounds like the `resp.Body` check may not be needed here:
https://stackoverflow.com/a/33238755/156674
https://golang.org/pkg/net/http/#Client.Do
> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close.

metal/eip_controlplane_reconciliation.go
```diff
@@ -106,9 +106,13 @@ func (m *controlPlaneEndpointManager) reconcileNodes(ctx context.Context, nodes
...
+		defer resp.Body.Close()
+	}
 	if err != nil || resp.StatusCode != http.StatusOK {
 		if err != nil {
-			klog.Errorf("http client error during healthcheck. err \"%s\"", err)
+			klog.Errorf("http client error during healthcheck, will try to reassign to a healthy node. err \"%s\"", err)
```
displague, line 115: `\"%s\"` could be `%q`

displague (approved):
the logging and defer changes make sense (I didn't look outside the box of this PR).

metal/eip_controlplane_reconciliation.go
deitch: You are right, but it doesn't hurt.

PR #173, opened 2021-05-21
configure EIP port and healthcheck ports from existing service with env var override

### LINKED ISSUE #171: The "cloud-provider-equinix-metal-kubernetes-external" service has an incorrect port
The "cloud-provider-equinix-metal-kubernetes-external" service that the CCM creates when you enable the elastic IP for control plane nodes feature creates a service on port 443, not 6443. The EIP healthcheck checks on port 6443, so it will fail unless you manually edit the service.

```
  ports:
  - name: https
    nodePort: 32373
    port: 443
    protocol: TCP
    targetPort: 6443
```

port: 443 should be changed to 6443 by default.

deitch:
Fixes #171 

Affects only control plane EIP assignment.

Before this change:

* EIP port always is set to 443
* Healthchecks on EIP port defaults to 6443, able to override via env var `METAL_API_SERVER_PORT` (which always has to be set to match 443)
* Healthchecks on control plane nodes always check on port 6443, no matter what the kube-apiserver actually is listening on

After this change:

* EIP port is set via env var `METAL_API_SERVER_PORT`; if it is not set, or is `0`, use the same port as the `kube-apiserver`
* Healthchecks on EIP port match the previous line
* Healthchecks on control plane nodes are set to whatever port the `kube-apiserver` is listening on, determined by checking the Kubernetes service `default/kubernetes`

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ This section lists each configuration option, and whether it can be set by each 
 | Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
 | Kubernetes annotation to set BGP MD5 password, base64-encoded (see security warning below) |   | `METAL_ANNOTATION_BGP_PASS` | `annotationBGPPass` | `"metal.equinix.com/bgp-pass"` |
 | Tag for control plane Elastic IP |    | `METAL_EIP_TAG` | `eipTag` | No control plane Elastic IP |
-| Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | `6443` |
+| Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | Same as `kube-apiserver` on control plane nodes, same as `0` |
 | Filter for cluster nodes on which to enable BGP |    | `METAL_BGP_NODE_SELECTOR` | `bgpNodeSelector` | All nodes |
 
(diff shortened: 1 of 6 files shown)


deitch:
Full disclosure: this only has been _partially_ tested. I won't have time to do a full set of testing until next week.

deitch:
This one is good to go. Can I get a review on it please?

README.md
```diff
@@ -418,7 +418,9 @@ To enable CCM to manage the control plane EIP:
 
 1. Create an Elastic IP, using the Equinix Metal API, Web UI or CLI
 1. Put an arbitrary but unique tag on the EIP
-1. When starting the CCM, set the [configuration][Configuration] for the control plane EIP tag, e.g. env var `METAL_EIP_TAG=<tag>`, where `<tag>` is whatever tag you set on the EIP
+1. When starting the CCM
+   * set the [configuration][Configuration] for the control plane EIP tag, e.g. env var `METAL_EIP_TAG=<tag>`, where `<tag>` is whatever tag you set on the EIP
+   * (optional) set the port that the EIP should listen o; by default, it will be the same port as the `kube-apiserver` on the control plane nodes, but you can change it with `METAL_API_SERVER_PORT=<port>`
```
displague, line 423:
```suggestion
   * (optional) set the port that the EIP should listen on; by default, or when set to `0`, it will use the same port as the `kube-apiserver` on the control plane nodes. This port can also be specified with `METAL_API_SERVER_PORT=<port>`
```

deitch:
Took your recommended change @displague ; let's get this in?

thebsdbox (approved):
Looks OK to me.

PR #175, opened 2021-06-02
exempt controlplane Service from loadbalancer management

### LINKED ISSUE #174: When MetalLB is enabled, BGP messes up the controlplane node
@deitch

deitch:
Fixes #174 

Ensures that the special `Service` for the control plane EIP does not try to get managed by the load balancer.


thebsdbox (approved):
Looks OK to me.

PR #180, opened 2021-06-22
Add permission to read and update events

enkelprifti98:
Permission to read and update events is needed for the CCM otherwise logs will show errors trying to read events when a new  LoadBalancer service has been created.

### DIFF
--- a/deploy/chart/templates/clusterrole.yaml
+++ b/deploy/chart/templates/clusterrole.yaml
@@ -70,4 +70,15 @@ rules:
       - watch
       - update
       - patch
+  - apiGroups:
+      - ''
+    resources:
+      - events
+    verbs:
+      - create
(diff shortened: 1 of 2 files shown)


deitch (changes requested):
Do you mind updating the helm chart as well? Just this file [here](https://github.com/equinix/cloud-provider-equinix-metal/blob/master/deploy/chart/templates/clusterrole.yaml), shouldn't be difficult.

enkelprifti98:
> Do you mind updating the helm chart as well? Just this file [here](https://github.com/equinix/cloud-provider-equinix-metal/blob/master/deploy/chart/templates/clusterrole.yaml), shouldn't be difficult.

Done

deitch (approved):

PR #185, opened 2021-06-30
Implement Instances v2

### LINKED ISSUE #133: Logs spammed with "The Packet cloud provider does not support InstancesV2"
It seems running CCM from image `docker.io/packethost/packet-ccm:bba111cc65391734bb78dbaf99ed2f56a9a1463d` produces a lot of spammy logs like shown below:
```
W0125 09:42:37.963982       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:37 [DEBUG] GET https://api.equinix.com/metal/v1/devices/aaf4679f-13f4-46c2-8519-ddbc4184438e?include=facility
W0125 09:42:38.138011       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:38 [DEBUG] GET https://api.equinix.com/metal/v1/devices/aaf4679f-13f4-46c2-8519-ddbc4184438e?include=facility
W0125 09:42:38.321460       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:38 [DEBUG] GET https://api.equinix.com/metal/v1/devices/8af3e9af-e1cd-4311-9c53-740a5c2e470b?include=facility
W0125 09:42:38.512576       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:38 [DEBUG] GET https://api.equinix.com/metal/v1/devices/8af3e9af-e1cd-4311-9c53-740a5c2e470b?include=facility
W0125 09:42:38.687375       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:38 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:42:38.864293       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:38 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:42:44.038321       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:44 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:42:44.220514       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:44 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:42:49.403057       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:49 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:42:49.577594       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:42:49 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
W0125 09:46:31.563386       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:46:31 [DEBUG] GET https://api.equinix.com/metal/v1/devices/aaf4679f-13f4-46c2-8519-ddbc4184438e?include=facility
W0125 09:46:32.027907       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:46:32 [DEBUG] GET https://api.equinix.com/metal/v1/devices/8af3e9af-e1cd-4311-9c53-740a5c2e470b?include=facility
W0125 09:46:32.238768       1 cloud.go:154] The Packet cloud provider does not support InstancesV2
2021/01/25 09:46:32 [DEBUG] GET https://api.equinix.com/metal/v1/devices/ba5666bd-3c48-498b-bc9c-bc2e288900d4?include=facility
```

Is this something which could be silenced?

### LINKED ISSUE #164: define zone topology using EM metro
Assuming Metros (https://metal.canny.io/changelog/new-metros-feature-live) fits the categorical description of Kubernetes Topology Zones,  https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesiozone, the CCM should annotation nodes with the Equinix Metal "metro" as the zone.

Every EM device instance (those created before Metros and after) have a metro field value.
* https://github.com/packethost/packngo/blob/master/devices.go#L57

displague:
Breaking Changes:
- `--provider-config` replaced with standard `--cloud-config` (defaults to `/etc/kubernetes/cloud.cfg`)
  - refactors main.go, moving much to metal/config.go

     This was done because `app.NewCloudControllerManagerCommand` has a new signature which ultimately passes a config file along.  I compared notes with the GCP, DigitalOcean, vSphere, and OpenStack cloud providers.

Changes:
- Updates k8s.io dependencies to 1.21.2 
  - Removed the complex `replace` lines and `GOMODULES.md` (see https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-505725449)
- disables the `routes` controller, do we need to disable others? Do we use the `cloud-node-lifecycle` controller?
- Fixes #133:
   Implements Instances v2 https://github.com/kubernetes/cloud-provider/blob/v0.21.2/cloud.go#L196-L212
- Fixes #164:
  Zone / Region mapped to Facility / Metro respectively, https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
  * Should we have concerns about existing project networks? (If nodes running across sjc1 and sv15 before the introduction of Metros were running the CCM, would/could they have been in the same cluster? Now they can.)
  * Can we ignore VLANs created before Metros were introduced because they were not supported by the CCM?
- updates packngo to 0.20.0
  - retryablehttp is no longer used by default. See https://github.com/equinix/terraform-provider-metal/blob/main/metal/config.go#L75-L94 if we want this behavior.


deitch:
The diffs were too big here; I think I need to just look at what it does.

Do you want to squash the commits?

deitch:
> Updates k8s.io dependencies to 1.21.2

As a general rule, aren't these backwards and forwards compatible one minor version? If this means ccm won't work with k8s <1.20 (i.e. 1.19, 1.18), that might be an issue. At the very least, we would need to cut a release with a clear warning that this won't work with those earlier ones. But that still might lead to some user pushback.

> Removed the complex replace lines and GOMODULES.

My oh my, yes!!

> If nodes running across sjc1 and sv15 before the introduction of Metros were running the CCM, would/could they have been in the same cluster? Now they can

They couldn't before. It was restricted to one facility. It discovered the facility from metadata where the ccm is running or override via env var. I wouldn't worry about users who managed somehow to bypass that.

But this does mean we need to rename the env var and how it is used.

> Can we ignore VLANs created before Metros were introduced because they were not supported by the CCM?

Yes. I wouldn't worry about it.

> Zone / Region mapped to Facility / Metro respectively. Does this fit the Kubernetes expectations around topology? Nodes in the same metro (region) will be able to rebalance pods and the EM networking will allow it.

My opinion is that it does, but I would like weigh-in from you and maybe @detiber as well. I am not as confident as I should be about it.

> disables the routes controller, do we need to disable others? Do we use the cloud-node-lifecycle controller?

It already was disabled by returning a `nil`?

deitch:
To enable `InstancesV2()`, don't you need to return the `&instances` [here](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60R161-R164) instead of `nil`?

deitch:
I have two structural comments.

### What actually calls the provider to register?

In the existing one, you call [metal.InitializeProvider(config)](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L74-L77), which, in turn, eventually calls [	cloudprovider.RegisterCloudProvider](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60L91).

In the change, that functionality is moved to [metal.init()](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60R81) which calls the same [cloudprovider.RegisterCloudProvider](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60R82). 

Is this a better structure, having it called by default? Why is that better than having the single entry point in the program know and call it? Or is it required by the new cloud-provider structure from k8s?

### creating the config

I am ambivalent about moving all of the logic to create the finalized config from `package main` to `package metal`. and further having it then read env vars and such. I like the clean idea that all interaction with _how_ this program started is in main (or perhaps a dedicated package), and by the time we call `init()` (or `InitializeProvider`), or its callback routine, we have figured all of that out.

Is this something semi-official?

main.go
```diff
@@ -1,235 +1,75 @@
...
-var (
-	providerConfig string
-)
-
 func main() {
 	rand.Seed(time.Now().UTC().UnixNano())
 
-	command := app.NewCloudControllerManagerCommand()
+	opts, err := options.NewCloudControllerManagerOptions()
+	opts.KubeCloudShared.CloudProvider.Name = metal.ProviderName
```
deitch, line 29: This probably should be checked for various `nil`s, as `KubeCloudShared` and `CloudProvider` are pointers.

displague: I copied these two lines from DO. Others may have it too.  I exported ProviderName to match the pattern.

displague:
> The diffs were too big here; I think I need to just look at what it does.
> Do you want to squash the commits?

The commits are logical, let me know if they make more sense (are smaller) viewed individually.

The commit that updates to 1.21.2 is separate from the one that introduces the `--cloud-config` handling. It is worth noting that the 1.21.2 change by itself will not compile because of the `app.NewCloudControllerManagerCommand` changes needed and added in the following commit.

 `--cloud-config` handling was mostly a move of the existing config functions from main.go to metal/config.go, changing metalConfig to read from a buffer instead of a filename.  I got lost in the pflag changes, I don't know what the original code was trying to accomplish. I see similar pflag lines in the DO, openstack, and vsphere providers, but I don't see evidence of those providers setting the config file via pflags.

displague:
> Is this a better structure, having it called by default? Why is that better than having the single entry point in the program know and call it? Or is it required by the new cloud-provider structure from k8s?

I think main.go's job is to say, "I have a plugin - here it is", the plugin is not really activated.

When the plugin is activated, it is handed a config buffer, perhaps there is some file permission juggling going on? (/etc/kubernetes/cloud.cfg is needed, but the ccm shouldn't have direct filesystem access to it?). That's the sense I get from the structure, but I couldn't find text to back this up.

https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60R84

> Is this something semi-official?

This is the pattern I found in the other providers I compared. The cloud-provider sample / example in the k8s repo sets this precedent.

displague:
> Is this a better structure, having it called by default? Why is that better than having the single entry point in the program know and call it? Or is it required by the new cloud-provider structure from k8s?

I think main.go's job is to say, "I have a plugin - here it is", the plugin is not really activated.

When the plugin is activated, it is handed a config buffer, perhaps there is some file permission juggling going on? (/etc/kubernetes/cloud.cfg is needed, but the ccm shouldn't have direct filesystem access to it?). That's the sense I get from the structure, but I couldn't find text to back this up.

https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-d677710ef260f87ecc6eedb2d9bf090758eaa9aeb4fd917444ff83ade963ca60R84

> Is this something semi-official?

This is the pattern I found in the other providers I compared. The cloud-provider sample / example in the k8s repo sets this precedent.

main.go
deitch: Heh, so did I (and many others) when getting started here. 

Either way, we should be careful and check for nils.

deitch: I dug into their (k8s) code, looks like this just is the way it is. But let's check those nils.

deitch:
> I think main.go's job is to say, "I have a plugin - here it is", the plugin is not really activated.

I see what it is doing now (took too much digging into the cloud-provider reference code). I don't love their approach, but I think what you are doing here is correct. I still am not madly in love with the env var overrides in the actual `pkg/metal`, but they don't seem to consider the possibility of overrides. We will live with it.


> cloud-config

So does `--cloud-config` replace `--provider-config`? And it just creates an `io.Reader` to the file that is passed to the register function? Not sure why they changed that. Does it have any other impacts? Looking at the `k8s.io/cloud-provider` source code, it looks like it just passes it in as is.

If so, why do we need the `--provider-config` at all? We can deprecate it.

displague:
> So does --cloud-config replace --provider-config? And it just creates an io.Reader to the file that is passed to the register function? Not sure why they changed that.

Was provider-config ever standard? Perhaps it was conventional.

> Does it have any other impacts? Looking at the k8s.io/cloud-provider source code, it looks like it just passes it in as-is.
> If so, why do we need the --provider-config at all? We can deprecate it.

This PR removes --provider-config support. If we want to continue to support it, and deprecate it, I'll have to revisit the pflags code. After spending sometime in ~packet-cli~ metal-cli, I might be prepared for that.

deitch:
> Was provider-config ever standard? Perhaps it was conventional.

No idea. We add it explicitly to `main.go`, so if this is what "all the cool kids" are doing, I am fine just replacing it.

> The commit that updates to 1.21.2 is separate from the one that introduces the --cloud-config handling. It is worth noting that the 1.21.2 change by itself will not compile because of the app.NewCloudControllerManagerCommand changes needed and added in the following commit.

I don't think we want commits at any point that are not rebuildable on their own.

deitch:
Did you have a chance to address some of the comments, e.g. instances and instances v2 and what needs to be returned?

Also, if we are replacing `--provider-config` with `--cloud-config`, then we need to update the deployment yaml and helm chart.

I have some other work I want to do on CPEM, so once this is in a good state and we get it in, I will attack those.

deitch (dismissed):
LGTM. I am going to run some more tests, and then we can merge it in.

deitch:
@detiber and @displague where do we stand on this? I have not run full manual test yet, waiting to see if we are ready. It looks good to me, but there were some comments. Kindly comment here if we are ready, short of manual testing.

metal/cloud.go
```diff
@@ -42,6 +42,7 @@ type cloudService interface {
 
 type cloudInstances interface {
 	cloudprovider.Instances
```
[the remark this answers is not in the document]
displague: I think this might be needed for existing installs on older k8s versions to be able to update to this version of CPEM.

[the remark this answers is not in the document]
detiber: > I think this might be needed for existing installs on older k8s versions to be able to update to this version of CPEM.

I don't think that is the case, as far as I can tell the use (or not) of Instances/Instancesv2 is isolated to the controller manager code that we are importing from k8s.io/cloud-provider, initializing and running from within the binary that we ship. All use of the Instances/Instancesv2 interfaces appears to done at runtime and the instantiated types don't appear to be persisted anywhere.

Since the instantiated types are not used by external code or persisted anywhere and I'm not seeing any options to override the behavior and force the use of Instances over Instancesv2 when Instancesv2 is implemented, it appears that keeping around support for Instances would just leave around vestigial and unused code.

metal/cloud.go
```diff
@@ -78,22 +79,28 @@ func newCloud(metalConfig Config, client *packngo.Client) (cloudprovider.Interfa
...
-	client := packngo.NewClientWithAuth("", metalConfig.AuthToken, nil)
-	client.UserAgent = fmt.Sprintf("cloud-provider-equinix-metal/%s %s", VERSION, client.UserAgent)
-	cloud, err := newCloud(metalConfig, client)
-	if err != nil {
-		return fmt.Errorf("failed to create new cloud handler: %v", err)
-	}
-
-	// finally, register
-	cloudprovider.RegisterCloudProvider(providerName, func(config io.Reader) (cloudprovider.Interface, error) {
+func init() {
```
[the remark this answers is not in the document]
detiber: One of these days it'll bug me enough to start addressing some of these Go anti-patterns in the k8s code base and documentation, but today is not one of those days 😂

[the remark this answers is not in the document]
deitch: So this part is good for now, then?

metal/cloud.go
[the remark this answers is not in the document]
deitch: Agreed. @displague do you mind ripping it out? I can run tests against slightly older and slightly newer clusters, if need be.

deitch:
Looks like just that vestigial code, then?

metal/cloud.go
```diff
@@ -78,22 +79,28 @@ func newCloud(metalConfig Config, client *packngo.Client) (cloudprovider.Interfa
...
+	cloudprovider.RegisterCloudProvider(ProviderName, func(config io.Reader) (cloudprovider.Interface, error) {
 		// by the time we get here, there is no error, as it would have been handled earlier
+		metalConfig, err := getMetalConfig(config)
+		// register the provider
+		if err != nil {
+			return nil, fmt.Errorf("provider config error: %v", err)
+		}
+
+		// report the config
+		printMetalConfig(metalConfig)
```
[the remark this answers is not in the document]
displague: https://github.com/equinix/cloud-provider-equinix-metal/issues/213

displague:
@deitch  This has been rebased.  The tests have been updated for the region/zone change in naming.
* `regions` are now equated to Equinix Metal metros
* `zones` (also called `failureDomain` in the CloudProvide Zone API) are now equated to Equinix Metal facilities

Tests were also updated for instancev1 removal, updated with their instancesv2 equivalent data access methods.

Tests were also updated because packngo will now error if the UUID used in a request is invalid (we can not use equinixmetal://123 and expect the same error).

deitch:
@displague all of CI passed. Does that mean this is good to go?

deploy/template/deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
           - "--cloud-provider=equinixmetal"
           - "--leader-elect=false"
           - "--authentication-skip-lookup=true"
-          - "--provider-config=/etc/cloud-sa/cloud-sa.json"
+          - "--cloud-config=/etc/cloud-sa/cloud-sa.json"
```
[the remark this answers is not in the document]
deitch: See #217

deitch:
I am going to do one more run of testing, and then going to merge it in.

deitch:
My testing has failed. I think I know why. 

When a node is created with the cloud-provider external, `k8s.io/cloud-provider` initializes the node, which makes sense. This happens in [cloud.syncNode()](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L374).

As part of that process, it [gets the device providerID](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L394) from the CPEM, passing that ID to the CPEM when it [gets instance metadata](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L399).

However, the path diverges between those two steps. When it [gets the device providerID](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L394) it calls [cnc.getproviderID()](https://github.com/kubernetes/cloud-provider/blob/5675546fef8efee38ac165cab3a23adf8677cb30/controllers/node/node_controller.go#L530), which checks for InstancesV2, and if it does, returns no provider ID but no error, see [here](https://github.com/kubernetes/cloud-provider/blob/5675546fef8efee38ac165cab3a23adf8677cb30/controllers/node/node_controller.go#L535-L538):

```go
	if _, ok := cnc.cloud.InstancesV2(); ok {
		// We don't need providerID when we call InstanceMetadata for InstancesV2
		return "", nil
	}
```

So when it later [gets instance metadata](https://github.com/kubernetes/cloud-provider/blob/master/controllers/node/node_controller.go#L399), it passes `""` as the provider ID. An InstancesV2 implementation is expected to be able to get instance metadata without actually having the provider ID. Our CPEM relies heavily on the provider ID to extract the, well, provider UUID to use the EQXM API to get the metadata (and everything else).

We need to determine how InstancesV2 is expected to get metadata without providerID and retrieve it. Your code has [InstanceMetadata() here](https://github.com/equinix/cloud-provider-equinix-metal/pull/185/files#diff-3ddaf4333cb941b353834040ab733477c56a207db5bcf2e854b62fe663db5494R208).

deitch:
FWIW, the AWS provider has [InstanceMetadata](https://github.com/kubernetes/cloud-provider-aws/blob/master/pkg/providers/v2/instances.go#L111) (and every other func that uses it) call [getInstance](https://github.com/kubernetes/cloud-provider-aws/blob/488b2151726d917803107e440dd23af6661c9498/pkg/providers/v2/instances.go#L137-L158) which checks if a provider ID is available or not, and figure it out from there based on DNS name.

We could do something similar, just have a `deviceFromNode()` which checks (pseudo-code):

* if `node.providerID` set, call `deviceFromProviderID(node.providerID)`
* if not, call `deviceFromName(node.Name)`

displague:
@deitch from pseudo to soo done, I hope. I added additional tests and revised the existing test error expectations around "providerID cannot be empty". ProviderID can be empty, and the node name (object metadata name, which is required) will be used.

displague:
@deitch in the future, I can see us adding a local metadata test (metadata.platformequinix.com) to deviceByNode. I would imagine this would fit between the provider id and node name lookups.

deitch:
> local metadata test (metadata.platformequinix.com) to deviceByNode

We are trying to get rid of local metadata entirely from CPEM. One cannot fairly depend on the CPEM running in the same facility as the worker nodes (or even on EQXM at all).

metal/cloud.go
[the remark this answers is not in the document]
displague: closing this thread since Instances v1 has been ripped out.  There may be some dead code paths that I didn't catch, and some of the node metadata lookup tests may be redundant (as a result of old tests being refactored to fit the new patterns). I think we can clean that up in the future.

README.md
```diff
@@ -195,7 +195,7 @@ This section lists each configuration option, and whether it can be set by each
 
 | Purpose | CLI Flag | Env Var | Secret Field | Default |
 | --- | --- | --- | --- | --- |
-| Path to config secret |    |    | `provider-config` | error |
+| Path to config secret |    |    | `cloud-config` | error |
 | API Key |    | `METAL_API_KEY` | `apiKey` | error |
 | Project ID |    | `METAL_PROJECT_ID` | `projectID` | error |
 | Facility |    | `METAL_FACILITY_NAME` | `facility` | read metadata on host on which CCM is running, else error |
```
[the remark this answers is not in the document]
displague: #214 added support for using the metro for elastic IPs.

This PR addresses the concerns of #164 by defining the cloud provider Zone response.

metal/devices.go
```diff
@@ -188,6 +193,66 @@ func (i *instances) InstanceShutdownByProviderID(_ context.Context, providerID s
...
 }
 
+// InstanceShutdown returns true if the node is shutdown in cloudprovider
+func (i *instances) InstanceShutdown(ctx context.Context, node *v1.Node) (bool, error) {
+	return i.InstanceShutdownByProviderID(ctx, node.Spec.ProviderID)
+}
+
+// InstanceExists returns true if the node exists in cloudprovider
+func (i *instances) InstanceExists(ctx context.Context, node *v1.Node) (bool, error) {
+	return i.InstanceExistsByProviderID(ctx, node.Spec.ProviderID)
```
deitch, line 203: I really would like to have this be more flexible, using your new `deviceByNode()`, so it can handle with or without a provider ID. But as [InstancesV2](https://pkg.go.dev/k8s.io/cloud-provider#InstancesV2) says that `InstanceExists` and `InstanceShutdown` could use either, let's leave this. I will open a new PR to do that once this is in.

deitch:
Retesting.

deitch:
So, it _mostly_ works, but the bgp does not. The reason for that is that there are other places that rely on the node having a provider ID set. Yet, even after instance metadata is set correctly during initialization, it still does not set the provider ID.

Logs from setting annotations:

```
I1130 17:35:57.728210   17585 node_controller.go:492] Adding node label from cloud provider: beta.kubernetes.io/instance-type=c3.medium.x86
I1130 17:35:57.728239   17585 node_controller.go:493] Adding node label from cloud provider: node.kubernetes.io/instance-type=c3.medium.x86
I1130 17:35:57.728260   17585 node_controller.go:504] Adding node label from cloud provider: failure-domain.beta.kubernetes.io/zone=da11
I1130 17:35:57.728273   17585 node_controller.go:505] Adding node label from cloud provider: topology.kubernetes.io/zone=da11
I1130 17:35:57.728308   17585 node_controller.go:515] Adding node label from cloud provider: failure-domain.beta.kubernetes.io/region=da
I1130 17:35:57.728324   17585 node_controller.go:516] Adding node label from cloud provider: topology.kubernetes.io/region=da
I1130 17:35:57.728334   17585 cloud.go:190] called ProviderName, returning equinixmetal
```

Notice that it called the `ProviderName()`. And here is the node:

```yml
apiVersion: v1
kind: Node
metadata:
  annotations:
    kubeadm.alpha.kubernetes.io/cri-socket: /var/run/dockershim.sock
    node.alpha.kubernetes.io/ttl: "0"
    volumes.kubernetes.io/controller-managed-attach-detach: "true"
  creationTimestamp: "2021-11-30T09:20:20Z"
  labels:
    beta.kubernetes.io/arch: amd64
    beta.kubernetes.io/instance-type: c3.medium.x86
    beta.kubernetes.io/os: linux
    failure-domain.beta.kubernetes.io/region: da
    failure-domain.beta.kubernetes.io/zone: da11
    kubernetes.io/arch: amd64
    kubernetes.io/hostname: k8s-master
    kubernetes.io/os: linux
    node-role.kubernetes.io/control-plane: ""
    node-role.kubernetes.io/master: ""
    node.kubernetes.io/exclude-from-external-load-balancers: ""
    node.kubernetes.io/instance-type: c3.medium.x86
    topology.kubernetes.io/region: da
    topology.kubernetes.io/zone: da11
spec:
  taints:
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
```

I left out the irrelevant status and managedFields. I am going to dig further, but does InstancesV2 _not_ set the provider ID?

metal/devices.go
```diff
@@ -188,6 +193,66 @@ func (i *instances) InstanceShutdownByProviderID(_ context.Context, providerID s
...
+
+	if device.Facility != nil {
+		z = device.Facility.Code
+	}
+	if device.Metro != nil {
+		r = device.Metro.Code
+	}
+
+	return &cloudprovider.InstanceMetadata{
+		ProviderID:    node.Spec.ProviderID,
```
deitch, line 240: Aha, here! This is where you are expected to set the provider ID, but it is just passing back what it received.

I believe this should be `fmt.Sprintf("%s://%s", ProviderName, device.ID)`

deitch (changes requested):
One missing piece that I found, which is causing the rest to fail.

metal/devices.go
deitch: Confirmed that the above works.

deitch, line 240:
```suggestion
		ProviderID:    fmt.Sprintf("%s://%s", ProviderName, device.ID),
```

deitch (changes requested):

metal/devices.go
displague: @deitch I just want to say, VSCode Copilot is amazing.. 

image

It's like a portable @deitch.

deitch: haha! Only as portable as your laptop. 😁 

Make the changes, I can approve and merge this in. Already fully tested.

displague: pushed

deitch (approved):

deitch:
Once CI is clean, we can merge it in

displague:
I raced your approval with a rebase.  
I squashed out the commit that introduced packngo 0.17. This commit was appearing after the packngo 0.20 commit and it was only removing commented out code.

I'll click merge when this last test finishes, thanks for the diligent review and exceptional patience!

displague:
I misspoke - the 'build' action was still processing when I rebased. Yipes, that job takes some time doing package installs.

deitch:
Yeah, it is really slow. Long story. I will take a look at it once this is in.

deitch:
Thank you for all the great work on this @displague !

PR #187, opened 2021-07-14
add annotation for private network

deitch:
In addition to the BGP annotations we already use. This came as an outgrowth of Gardener work, but I started seeing other potential use cases. In any case, the right place to do this is CPEM.

@displague I know this will cause you to have to rebase your big PR; sorry about that. this was getting in the way of some other work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,7 @@ This section lists each configuration option, and whether it can be set by each 
 | Kubernetes annotation to set BGP peer's IPs |   | `METAL_ANNOTATION_PEER_IPS` | `annotationPeerIPs` | `"metal.equinix.com/peer-ip"` |
 | Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
 | Kubernetes annotation to set BGP MD5 password, base64-encoded (see security warning below) |   | `METAL_ANNOTATION_BGP_PASS` | `annotationBGPPass` | `"metal.equinix.com/bgp-pass"` |
+| Kubernetes annotation to set the CIDR for the network range of the private address |  | `METAL_ANNOTATION_NETWORK_IPV4_ PRIVATE` |  `annotationNetworkIPv4Private` | `metal.equinix.com/network/4/private` |
 | Tag for control plane Elastic IP |    | `METAL_EIP_TAG` | `eipTag` | No control plane Elastic IP |
 | Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | Same as `kube-apiserver` on control plane nodes, same as `0` |
 | Filter for cluster nodes on which to enable BGP |    | `METAL_BGP_NODE_SELECTOR` | `bgpNodeSelector` | All nodes |
@@ -524,13 +525,15 @@ _not_ recommended to override them. However, you can do so, using the options in
 Set of servers on which BGP will be enabled can be filtered as well, using the the options in [Configuration][Configuration].
 Value for node selector should be a valid Kubernetes label selector (e.g. key1=value1,key2=value2).
 
-In addition to enabling BGP and setting ASNs, the Equinix Metal CCM sets Kubernetes annotations on each cluster node.
-It sets annotations for:
+## Node Annotations
+
+The Equinix Metal CCM sets Kubernetes annotations on each cluster node:
 
 * Node, or local, ASN, default annotation `metal.equinix.com/node-asn`
 * Peer ASNs, comma-separated if multiple, default annotation `metal.equinix.com/peer-asns`
 * Peer IPs, comma-separated if multiple, default annotation `metal.equinix.com/peer-ips`
 * Source IP to use when communicating with upstream peers, default annotation `metal.equinix.com/src-ip`
+* CIDR of the private network range in the project which this node is part of, default annotation `metal.equinix.com/network-ipv4/private`
 
 These annotation names can be overridden, if you so choose, using the options in [Configuration][Configuration].
 
--- a/main.go
+++ b/main.go
@@ -23,21 +23,22 @@ import (
 )
 
 const (
-	apiKeyName                   = "METAL_API_KEY"
-	projectIDName                = "METAL_PROJECT_ID"
-	facilityName                 = "METAL_FACILITY_NAME"
-	loadBalancerSettingName      = "METAL_LOAD_BALANCER"
-	envVarLocalASN               = "METAL_LOCAL_ASN"
-	envVarBGPPass                = "METAL_BGP_PASS"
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -208,6 +208,7 @@ This section lists each configuration option, and whether it can be set by each
 | Kubernetes annotation to set BGP peer's IPs |   | `METAL_ANNOTATION_PEER_IPS` | `annotationPeerIPs` | `"metal.equinix.com/peer-ip"` |
 | Kubernetes annotation to set source IP for BGP peering |   | `METAL_ANNOTATION_SRC_IP` | `annotationSrcIP` | `"metal.equinix.com/src-ip"` |
 | Kubernetes annotation to set BGP MD5 password, base64-encoded (see security warning below) |   | `METAL_ANNOTATION_BGP_PASS` | `annotationBGPPass` | `"metal.equinix.com/bgp-pass"` |
+| Kubernetes annotation to set the CIDR for the network range of the private address |  | `METAL_ANNOTATION_NETWORK_IPV4_ PRIVATE` |  `annotationNetworkIPv4Private` | `metal.equinix.com/network/4/private` |
```
displague, line 211:
```suggestion
| Kubernetes annotation to set the CIDR for the network range of the private address |  | `METAL_ANNOTATION_NETWORK_IPV4_PRIVATE` |  `annotationNetworkIPv4Private` | `metal.equinix.com/network/4/private` |
```

displague: Are these "private" VLAN addresses or 10/8 project-scoped management addresses?

I realize the terms are overlapping and fuzzy:
https://metal.equinix.com/developers/docs/networking/ip-addresses/

deitch: 10/8 project-scoped. What k8s also calls "private" or "internal"

deitch:
Took your doc suggestion. Anything else @displague ?

main.go
```diff
@@ -191,6 +192,12 @@ func getMetalConfig(providerConfig string) (metal.Config, error) {
 		config.AnnotationBGPPass = annotationBGPPass
 	}
 
+	config.AnnotationNetworkIPv4Private = metal.DefaultAnnotationNetworkIPv4Private
+	annotationNetworkIPv4Private := os.Getenv(envVarAnnotationIPv4NetworkPrivate)
```
displague, line 196: perhaps the env var constant could match the config name: envVarAnnotationNetworkIPv4Private (to make future code searches easier)

displague (approved):
This looks reasonable.  I pointed out a logger typo.

I wonder about the intention of this PR. It seems that this allows for a user-specified private range, by setting the annotation and value rather than looking one up. This would better accommodate VLAN configured networks.  Some documentation describing that use case (and how to accomplish it) may be helpful.

How is this annotation used?

How is this value different than the node InternalIP address?  Why are we not reflecting that address in this annotation?

metal/devices.go
```diff
@@ -252,3 +256,84 @@ func (i *instances) deviceFromProviderID(providerID string) (*packngo.Device, er
...
+					}
+				} else {
+					klog.V(2).Infof("instances.reconcileNodes(): no change to annotations for %s", node.Name)
+				}
+			}
+		}
+	case ModeRemove:
+		klog.V(2).Info("instances.reconcileNodes(): nothing to do for removing nodes")
+	}
+	klog.V(2).Info("instamces.reconcileNodes(): complete")
```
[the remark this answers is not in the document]
deitch: Nah, you never heard of `instamces`? AWS has ec2 "instances", and EQXM has "instamces", ("m" for "metal") 😆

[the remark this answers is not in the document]
deitch: Fixed

main.go
deitch: That's a good idea. Will do.

deitch: Done.

deitch:
>  I wonder about the intention of this PR. It seems that this allows for a user-specified private range, by setting the annotation and value rather than looking one up from the EM API. This would better accommodate VLAN configured networks where the EM API is unaware of the IP configuration within the VLAN. Some documentation describing that use case (and how to accomplish it) may be helpful.

That wasn't at all the intent here, see below. That having been said, I am open to new intents.

> How is this annotation used?

This annotation - like the others - convey information about the device that normally would be gotten via metadata (if on the node) or via the EQXM API. Actually, CPEM uses the EQXM API to get these and add them.

> How is this value different than the node InternalIP address? Why are we not reflecting that address (and range) in this annotation?

The internalIP address is the actual address of the node itself. This annotation gets the range in which the internalIP sits. kubelet already gets the private IP and adds it as a property of the node, so we don't need to annotate it.

Here is an example of a node which I currently am running:

```json
      {
        "id": "eb8d4b25-9780-4b53-a0b5-5d1ed99cca48",
        "address_family": 4,
        "netmask": "255.255.255.254",
        "created_at": "2021-06-14T14:37:48Z",
        "public": false,
        "cidr": 31,
        "management": true,
        "enabled": true,
        "network": "10.99.173.130",
        "address": "10.99.173.131",
        "gateway": "10.99.173.130",
        "parent_block": {
          "network": "10.99.173.128",
          "netmask": "255.255.255.128",
          "cidr": 25,
          "href": "/ips/e9753813-eec6-4b4a-8db9-5712fbbeda21"
        }
      }
```

When this node is created, the node's InternalIP will be `10.99.173.131/31`. But what private network range is it in? That is given by the `parent_block` (same on the API), which is `10.99.173.128/25`.

The annotation adds that.

Does that explain it?

displague:
> The internalIP address is the actual address of the node itself. This annotation gets the range in which the internalIP sits. kubelet already gets the private IP and adds it as a property of the node, so we don't need to annotate it.

These are not necessarily the same networks, which is why I think it would help to expand on the intent and usage.  The `InternalIP` could be a 192.168/16 member configured on a VLAN while the EM API reports a 10/25 network.

deitch:
Hmm that is a good point. 

The intended usage is where you need to know what the network range is. With other clouds, you actually know it because you set up a VPC and subnets. Here, that doesn't exist. 

What are you suggesting?

displague:
I've been wondering about what CPEM might look like in pure layer-2 clusters. Cloud-provide-kube-vip may be a better fit in that scenario. In a mixed cluster, with control plane nodes in hybrid mode and worker nodes in layer-2 we may have need for varied internal network configurations. 

The upcoming gateway feature opens up some configurations I'd like to explore.

deitch:
> I've been wondering about what CPEM might look like in pure layer-2 clusters. Cloud-provide-kube-vip may be a better fit in that scenario. In a mixed cluster, with control plane nodes in hybrid mode and worker nodes in layer-2 we may have need for varied internal network configurations.
>
> The upcoming gateway feature opens up some configurations I'd like to explore.

This does open up some interesting new directions. I will hit you up on slack, we can arrange some time to set up a discussion, including others who know. 

In the meantime, since that will be a material rework, can we move ahead with this one?

displague (approved):

PR #195, opened 2021-08-10
Add missing operator: Exists to Deployment

flokli:
The default value for `operator` is `Equal`, but this doesn't provide
any value to check against, causing this to not be scheduled on control
plane nodes with the following taint:

```
  - effect: NoSchedule
    key: node-role.kubernetes.io/master
    value: "true"
```

Setting the operator to `Exists` will cause this to be scheduled.

cc @rawkode


deitch (approved):

flokli:
@deitch can we get a release containing this?

deitch:
Sure, no reason not to have a patch release for this bug fix.

Issue #44, opened 2020-05-01
Add option to use a specific range of IPs


enkelprifti98:
This is a necessary use case for customers that bring their own IPs, or Packet global anycast IPs.

displague:
We’ve been using kube-vip in the latest integrations (anthos in this case), metallb is not used at all.

We don’t have an annotation that the CCM can recognize to associate addresses to label in this way, that I know of.

The CCM will provision an extra IP if needed, but there is no way to request that the created IP be tagged with a tag or request an IP from reservations with a given tag.  I think a CCM recognized annotation that sets up this tag dependent behavior would allow us to ensure that multiple services claim the same address without knowing the address in advance (which is what spec.loadBalancerIP allows for today).

deitch:
We may actually be part of the way there. The following is paraphrased from the README, in both the kube-vip and metallb sections.

For each service of `type=LoadBalancer` currently in the cluster or added:

* if an Elastic IP address reservation with the appropriate tags exists, and the Service already has that IP address affiliated with it, it is ready; ignore
* if an Elastic IP address reservation with the appropriate tags exists, and the Service does not have that IP affiliated with it, add it to the service spec (if configured for metallb, also ensure it is in the pools of the metallb ConfigMap with `auto-assign: false`)
* if an Elastic IP address reservation with the appropriate tags does not exist, create it and add it to the services spec (if configured for metallb, also ensure is in the pools of the metallb ConfigMap with `auto-assign: false`)

The question is, are those tags usable as is? Or do we need something else?

displague:
Thanks for pointing this out, @deitch. This comment will include a little thinking-out-loud as I go through the README.

> When starting the CCM, set the env var `METAL_EIP_TAG=<tag>`, where <tag> is whatever tag you set on the EIP

This tag becomes a CCM-scoped filter, I think we would benefit from an application-scoped setting, specified within each service (as an annotation or label). 

Let's say `METAL_EIP_TAG` was set. If two Service/LoadBalancer resources were created with the same tag annotation, and we need a new /32 Elastic IP reservation to fulfill the LoadBalancer request, I would imagine the CCM would request a new reservation, tagging that reservation with both the METAL_EIP_TAG and the tag identified in the annotation.  The second LoadBalancer resource would, barring race conditions, find that same address.

[...]

PR #196, opened 2021-08-20
document bring-your-own-IP

deitch:
This already worked, but never was documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![Kubernetes Cloud Provider for Equinix Metal](docs/images/github_kubernetes-cloud-provider_share.jpeg))
+![Kubernetes Cloud Provider for Equinix Metal](docs/images/github_kubernetes-cloud-provider_share.jpeg)
 
 # Kubernetes Cloud Controller Manager for Equinix Metal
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -250,10 +250,44 @@ When a load balancer is enabled, the CCM does the following:
...
+    - protocol: TCP
+      port: 80
+      targetPort: 9376
+  type: LoadBalancer
+  loadBalancerIP: 145.60.80.60
+```
+
+CCM will detect that `loadBalancerIP` already was set and not try to create a new Equinix Metal Elastic IP.
+
+If the `Service.Spec.LoadBalancerIP` was *not* set, then CCM will use the Equinix Metal API to request a new, facility-specific
```
displague, line 285: This reminds me that we should revisit https://github.com/equinix/cloud-provider-equinix-metal/issues/44#issuecomment-789817232

Do we provide any annotations today that change the behavior of _how_ we `request a new one`?

displague (approved):
Left a question - but looks good to me

README.md
deitch: We do not now. And, yes, we should revisit #44. But I wouldn't hold this up for that.

Issue #200, opened 2021-09-02
CCM reconciling and updating services despite loadbalancer setting `empty://`


ipochi:
When setting the `loadbalancer: "empty://"` , it is expected that CCM won't interfere with the loadbalancing and only concerns itself with bgp only.

In such a scenario, CCM should not start Node and Service reconcilers as mentioned [here](https://github.com/equinix/cloud-provider-equinix-metal/blob/master/metal/loadbalancers.go#L109-#L123).

Yet, what happens is that reconcilers are started and CCM interferes by updating the Services of type `LoadBalancer`

As an example, I installed MetalLB and installed Contour

As seen below, immediately MetalLB provides the load balancer IP to the envoy service 
```
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: contour
    meta.helm.sh/release-namespace: projectcontour
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
  creationTimestamp: "2021-09-02T20:43:09Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: envoy
  namespace: projectcontour
  resourceVersion: "1611458"
  uid: c9ed277f-2878-4feb-8040-a650945faff1
spec:
  clusterIP: 10.3.15.255
  clusterIPs:
  - 10.3.15.255
  externalTrafficPolicy: Local
  healthCheckNodePort: 31765
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    nodePort: 31911
    port: 80
    protocol: TCP
    targetPort: 8080
  - name: https
    nodePort: 31764
    port: 443
    protocol: TCP
    targetPort: 8443
  selector:
    app: envoy
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 10.1.1.1
   ```
   
   Moments later the service is modified by CCM, below is the updated yaml of service
   ```
   apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: contour
    meta.helm.sh/release-namespace: projectcontour
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
  creationTimestamp: "2021-09-02T20:43:09Z"
  labels:
    app.kubernetes.io/managed-by: Helm
  name: envoy
  namespace: projectcontour
  resourceVersion: "1611611"
[...]

PR #201, opened 2021-09-03
fix: dont start reconcilers when loadbalancer 'empty'

ipochi:
Fixes: #200

This PR addresses the issue where even if the `loadbalancer:
empty://` is provided, CCM reconciles the Node and Service Objects
and tries to set the IP addresses to the services of type `LoadBalancer`

This is noticed when CCM is deployed with `loadBalancer: empty://` and
MetalLB is installed.

When an example service is installed of type LoadBalancer
What happens is that, MetalLB correctly assigns an IP from the address
pool but CCM updates the service.

This is because, we disable the reconcilers based on the condition if
the implementor is `nil`, which in its current state is never nil for
`empty` as the impl is set to &empty.LB{}.

Signed-off-by: Imran Pochi <imran@kinvolk.io>

### DIFF
--- a/metal/loadbalancers.go
+++ b/metal/loadbalancers.go
@@ -8,7 +8,6 @@ import (
 	"net/url"
 
 	"github.com/equinix/cloud-provider-equinix-metal/metal/loadbalancers"
-	"github.com/equinix/cloud-provider-equinix-metal/metal/loadbalancers/empty"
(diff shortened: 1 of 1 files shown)


deitch:
As discussed on the [related issue](https://github.com/equinix/cloud-provider-equinix-metal/issues/200), this is correct behaviour per the documentation. If you want CCM to do nothing in this space, set `METAL_LOAD_BALANCER` to `""`.

PR #214, opened 2021-11-19
Add explicit metro support

deitch:
Explicitly uses an environment variable or config to set the metro, falling back to metadata. If both are specified, both are sent in the EIP creation request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,8 @@ This section lists each configuration option, and whether it can be set by each 
 | Path to config secret |    |    | `provider-config` | error |
 | API Key |    | `METAL_API_KEY` | `apiKey` | error |
 | Project ID |    | `METAL_PROJECT_ID` | `projectID` | error |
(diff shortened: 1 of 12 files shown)


thebsdbox (dismissed):
Looks OK, out of interest what happens when both a facility AND a metro are provided?

deitch:
It passes them both to the Equinix Metal API, so it defers to whatever the API does. I think right now, if they match, it uses it, else not. But I am not sure. The CPEM just doesn't care, trusts the API entirely.

Got to get CI green.

deitch:
OK, got CI. As we bumped the packngo library version, some of the calls were stricter on parameters, breaking some tests. Using properly formatted UUIDs works.

thebsdbox (approved):

deitch:
Finally ready for one last review @thebsdbox . I had to do some extensive testing, needed one more fix.

thebsdbox (approved):
`/LGTM`

PR #215, opened 2021-11-22
integration tests

deitch:
Create integration tests for CPEM. This is a draft PR, initially just working on the documentation as to what it should cover. Eventually this PR will include the actual code.


deitch:
This was supposed to be a draft PR; since GitHub doesn't appear to turn a real PR into a draft one (sort of like the arrow of time, only going forwards), I am closing this and opening a new draft one.

PR #220, opened 2021-12-14
Clean logging

deitch:
This has two commits.

The first restructures some of the files and funcs without changing any functionality. It just reduces duplicate code, makes it easier to understand, separates some things into their own file, etc.

The second cleans up the logging:

* do not return an error for something that isn't actually an error. If it is informative, log a message (at various log levels), but return `nil`
* if we have a non-fatal error in a loop, log the message, save the affected resource in a slice, and complete the loop. Only at the end should we return the error.

I have wanted to do this for a long time.

### DIFF
--- a/metal/eip_controlplane_reconciliation.go
+++ b/metal/eip_controlplane_reconciliation.go
@@ -88,7 +88,8 @@ func (m *controlPlaneEndpointManager) reconcileNodes(ctx context.Context, nodes 
 		m.inProcess = false
 	}()
 	if m.eipTag == "" {
-		return errors.New("control plane loadbalancer elastic ip tag is empty. Nothing to do")
+		klog.Info("control plane loadbalancer elastic ip tag is empty. Nothing to do")
+		return nil
 	}
 	ipList, _, err := m.ipResSvr.List(m.projectID, &packngo.ListOptions{
 		Includes: []string{"assignments"},
@@ -100,7 +101,7 @@ func (m *controlPlaneEndpointManager) reconcileNodes(ctx context.Context, nodes 
 	if controlPlaneEndpoint == nil {
 		// IP NOT FOUND nothing to do here.
 		klog.Errorf("elastic IP not found. Please verify you have one with the expected tag: %s", m.eipTag)
-		return err
+		return nil
 	}
 	if len(controlPlaneEndpoint.Assignments) > 1 {
 		return fmt.Errorf("the elastic ip %s has more than one node assigned to it and this is currently not supported. Fix it manually unassigning devices", controlPlaneEndpoint.ID)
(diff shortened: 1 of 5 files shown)


metal/eip_controlplane_reconciliation.go
```diff
@@ -224,7 +226,8 @@ func newControlPlaneEndpointManager(eipTag, projectID string, deviceIPSrv packng
 // the `default/kubernetes` service
 func (m *controlPlaneEndpointManager) reconcileServices(ctx context.Context, svcs []*v1.Service, mode UpdateMode) error {
 	if m.eipTag == "" {
-		return errors.New("elastic ip tag is empty. Nothing to do")
```
displague, line 227: I'm hitting this error in a cluster I just brought up under k8s 1.23, haven't looked into why yet.
Can you explain the change in behavior here?

displague: (why is this not an error)

deitch: Yeah. The way it is defined, if you set an EIP tag, then you want CPEM to manage it; if you do not set an EIP tag, then you are telling CPEM, "there is nothing for you to manage at the control plane EIP level." Lack of EIP tag = "nothing to do here". That definitely is _not_ an error, and never should have been treated as such.

deitch: Don't ask how many people that led astray. 🤦‍♂️

deitch:
Rebased.

metal/eip_controlplane_reconciliation.go
deitch: I sort of did `"Nothing to do."` as an indication that this is not an error, just nothing that needs to be done here. The correct answer to that is, "so don't make it an error!" 😬

metal/watchers.go
```diff
@@ -0,0 +1,143 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+)
+
+// createNodesWatcher create a cache.SharedIndexInformation for node handlers
+func createNodesWatcher(ctx context.Context, informer informers.SharedInformerFactory, cloudServices []cloudService) (cache.SharedIndexInformer, error) {
```
displague, line 16: If we were to implement `LoadBalancer`, we would get these node/service watchers/informers for free. 
I don't know if kube-vip or metallb register themselves as cloud-provider LoadBalancers (what does that really mean? I don't know.), I presume they do not.  If they do I would be curious to know what users would provider for Service loadBalancerClass https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class.

Why do we not implement LoadBalancer?

displague (approved):
Looks good to me - the errors that are returned vs logged make sense.

The watcher movement to a different file looks fine - perhaps we could move this to an internal package in the future or implement LoadBalancers ([see my comment](https://github.com/equinix/cloud-provider-equinix-metal/pull/220#discussion_r769842911)).

metal/watchers.go
deitch: We should. It is on my list of todos, to which I haven't gotten yet. Your work on `InstancesV2()` also inspired me to think about getting moving on that. If it simplifies our work, even better.

deitch: But, like moving `watchers.go` into an internal package, subject for another PR.

deitch:
CI failed because of `gofmt` issues. Fixed and pushed.

deitch:
CI green, ready for you @displague

displague (approved):

PR #221, opened 2021-12-14
report versions correctly

### LINKED ISSUE #142: Fix logged version formatting
Right now running version v3.0.0 logs the following version, which is incorrect:
```
I0128 21:37:43.995930       1 controllermanager.go:127] Version: v0.0.0-master+$Format:%h$
```

deitch:
Fixes #142 

Works with the Kubernetes standard versioning (without all of the branch complexity)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,14 +6,20 @@ PACKAGE_NAME?=github.com/equinix/cloud-provider-equinix-metal
 GIT_VERSION?=$(shell git log -1 --format="%h")
 VERSION?=$(GIT_VERSION)
 RELEASE_TAG ?= $(shell git tag --points-at HEAD)
+MOST_RECENT_RELEASE_TAG ?= $(shell git describe --abbrev=0  2>/dev/null || true)
+ifeq (,$(MOST_RECENT_RELEASE_TAG))
+MOST_RECENT_RELEASE_TAG = v0.0.0
+endif
 ifneq (,$(RELEASE_TAG))
-VERSION := $(RELEASE_TAG)-$(VERSION)
+VERSION := $(RELEASE_TAG)
+else
+VERSION := $(MOST_RECENT_RELEASE_TAG)-$(VERSION)
 endif
 GO_FILES := $(shell find . -type f -not -path './vendor/*' -name '*.go')
 BUILD_TAG ?= latest
 TAGGED_IMAGE ?= $(BUILD_IMAGE):$(BUILD_TAG)
 TAGGED_ARCH_IMAGE ?= $(TAGGED_IMAGE)-$(ARCH)
-LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)"'
+LDFLAGS ?= -ldflags '-extldflags "-static" -X "k8s.io/component-base/version.gitVersion=$(VERSION)"'
 
 # which arches can we support
 ARCHES=arm64 amd64
--- a/metal/cloud.go
+++ b/metal/cloud.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	cloudprovider "k8s.io/cloud-provider"
+	"k8s.io/component-base/version"
 	"k8s.io/klog/v2"
 )
 
@@ -96,7 +97,7 @@ func init() {
 
 		// set up our client and create the cloud interface
 		client := packngo.NewClientWithAuth("cloud-provider-equinix-metal", metalConfig.AuthToken, nil)
-		client.UserAgent = fmt.Sprintf("cloud-provider-equinix-metal/%s %s", VERSION, client.UserAgent)
+		client.UserAgent = fmt.Sprintf("cloud-provider-equinix-metal/%s %s", version.Get(), client.UserAgent)
 		cloud, err := newCloud(metalConfig, client)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create new cloud handler: %v", err)
--- a/metal/version.go
+++ b/metal/version.go
@@ -1,8 +1,0 @@
-// This file does not need to be updated. Release versions are set at build time (see LDFLAGS in the Makefile)
-
-package metal
-
-var (
-	// VERSION is reported in the API User-Agent
-	VERSION = "devel"
-)


Makefile
```diff
@@ -6,14 +6,20 @@ PACKAGE_NAME?=github.com/equinix/cloud-provider-equinix-metal
...
+VERSION := $(RELEASE_TAG)
+else
+VERSION := $(MOST_RECENT_RELEASE_TAG)-$(VERSION)
 endif
 GO_FILES := $(shell find . -type f -not -path './vendor/*' -name '*.go')
 BUILD_TAG ?= latest
 TAGGED_IMAGE ?= $(BUILD_IMAGE):$(BUILD_TAG)
 TAGGED_ARCH_IMAGE ?= $(TAGGED_IMAGE)-$(ARCH)
-LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)"'
+LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)" -X "k8s.io/component-base/version.gitVersion=$(VERSION)"'
```
displague, line 22:
```suggestion
LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/metal.VERSION=$(VERSION)" -X "k8s.io/component-base/version.gitVersion=$(VERSION)"'
```

displague: I've confirmed that the API is receiving UserAgents like this today (these are the most popular UAs for CPEM when creating BGP Sessions):
* `cloud-provider-equinix-metal/devel packngo/v0.17.0`
* `cloud-provider-equinix-metal/devel packngo/v0.19.2-0.20211119121357-05157869e685`
* `cloud-provider-equinix-metal/devel packngo/0.5.1`


Another approach to setting the version is to use the build time version:
```patch
diff --git a/metal/version.go b/metal/version.go
index e73560c..22022e2 100644
--- a/metal/version.go
+++ b/metal/version.go
@@ -2,7 +2,26 @@
 
 package metal
 
+import (
+	"runtime/debug"
+)
+
 var (
 	// VERSION is reported in the API User-Agent
 	VERSION = "devel"
 )
+
+func init() {
+	// look for the default version and replace it, if found, from runtime build info
+	if VERSION != "devel" {
+		return
+	}
+	bi, ok := debug.ReadBuildInfo()
+	if !ok {
+		return
+	}
+
+	// Version is set in artifacts built with -X github.com/equinix/cloud-provider-equinix-metal/metal.VERSION=1.2.3
+	// Ensure version is also set when installed via go install github.com/equinix/metal-cli/cmd/metal
+	VERSION = bi.Main.Version
+}
```

displague: I verified that your `Makefile` should work. I'm a bit confused by the format of the `--version` output ("Kubernetes ${CPEM version}" rather than "Kubernetes 1.23" or some multi-part version string "Kubernetes 1.23; CPEM 3.3.0")

```
go run -ldflags '-extldflags "-static" -X "github.com/equinix/cloud-provider-equinix-metal/metal.VERSION=5.6.7" -X "k8s.io/component-base/version.gitVersion=v4.3.2"' ./...  --version
```

Makefile
```diff
@@ -6,14 +6,20 @@ PACKAGE_NAME?=github.com/equinix/cloud-provider-equinix-metal
...
+VERSION := $(RELEASE_TAG)
+else
+VERSION := $(MOST_RECENT_RELEASE_TAG)-$(VERSION)
 endif
 GO_FILES := $(shell find . -type f -not -path './vendor/*' -name '*.go')
 BUILD_TAG ?= latest
 TAGGED_IMAGE ?= $(BUILD_IMAGE):$(BUILD_TAG)
 TAGGED_ARCH_IMAGE ?= $(TAGGED_IMAGE)-$(ARCH)
-LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/version.VERSION=$(VERSION)"'
+LDFLAGS ?= -ldflags '-extldflags "-static" -X "$(PACKAGE_NAME)/metal.VERSION=$(VERSION)" -X "k8s.io/component-base/version.gitVersion=$(VERSION)"'
```
deitch, line 22: This was a good catch, thank you @displague

deitch: I fixed it with your suggestion

Makefile
deitch: I do now actually know. The `--version` flag is part of the cloud-provider, not something we implement. 🤷‍♂️

deitch: You know, I think we should get rid of the `metal.VERSION=..` part . Why not just consume it from the same place that cloud-provider does?

I will do that. Hang tight.

deitch: OK, now it just uses the k8s version we set with `-X`.

For devel, down really care if it reports the default weird string of `v0.0.0-master+$Format:%h$`? I guess we could adopt your suggestion. It gets something useful out there.

It doesn't look like `debug.BuiltInfo` is very useful, see [here](https://go.dev/play/p/Z7cwPy3MzE_B)

deitch: Oh forget it. Let's just stick with the kubernetes default of "if you didn't set it, we will use that `v0.0.0-master...` string

displague: @deitch have a look at https://github.com/equinix/metal-cli/blob/main/cmd/metal/main.go and then `go install github.com/equinix/metal-cli/cmd/metal@latest` and `metal --version`

It doesn't show a tag in the playground because the playground code is not versioned :-)
It's fair that this can potentially return `(devel)` and therefore can not be used for the Kubernetes version `--version` wants (since that requires a `v` prefixed semver).

I think `version.Get` would return the CPEM version from the Makefile setting as a user-agent, this seems fine, but I think we want the previous approach (patched for metal.VERSION).

I suspect that the `--version` result should suggest the kubernetes version.


Unrelated, I noticed `rest "k8s.io/client-go/rest"``rest.DefaultKubernetesUserAgent()` while sifting through the gcp cloud provider for similar code. They don't use a Makefile though and I don't understand how bazel sets these ldflags: https://github.com/kubernetes/cloud-provider-gcp/blob/0aadb1237479f9b0975e884b26bac29cc9fca39b/defs/version.bzl.


---

It seems other cloud providers are not too worried about this result:

```
$ docker run digitalocean/digitalocean-cloud-controller-manager:v0.1.35  /bin/digitalocean-cloud-controller-manager --version
Kubernetes v0.0.0-master+$Format:%H$
```

```
$ docker run linode/linode-cloud-controller-manager:latest  --version
Linode Cloud Controller Manager starting up
SENTRY_DSN not set, not initializing Sentry
Kubernetes v0.0.0-master+$Format:%h$
```

```
$ docker run us.gcr.io/k8s-artifacts-prod/provider-aws/cloud-controller-manager:v1.22.0-alpha.0 --version
Kubernetes v0.0.0-master+$Format:%H$
```

```
$ docker run mcr.microsoft.com/oss/kubernetes/azure-cloud-controller-manager:v1.1.3 --version
Cloud Provider Azure v1.1.3
```

Found one! I think we can follow Azure's pattern ("Cloud Provider Equinix Metal v3.3.0") by defining a custom global `--version` flag.  I don't know if that would correct the logged version (as reported in #142).

deitch: > I suspect that the --version result should suggest the kubernetes version.

I don't think it really matters to anyone but us, but we should do something reasonable.

> by defining a custom global --version flag.

I have no issue with this. We might do it more simply than they did. 

>  I think we want the previous approach (patched for metal.VERSION).

The challenge is that the official `controllermanager.go` uses  `k8s.io/component-base/version.Get()`, which uses `k8s.io/component-base/version.gitVersion`, which is a private var and therefore cannot be set by our code, _only_ by `-X` flags. So either we override `--version` and the whole `Run()` entirely, or we just accept that ugly default of theirs.

We definitely do not want to reimplement the entire `Run()` in `controllermanager.go`!

Going to try and sum this up. 

1. We use version in three places:
   * `--version` - inherited from `cloud-provider` - acceptable to return either the string we set (see next section) or `Cloud Provider Equinix Metal <version string>`
   * log line - part of `Run()` in `cloud-provider` - returns just the `<version string>`
   * in user-agent - our code - `Cloud Provider Equinix Metal <version string>`
1. We want the `<version string>` to be:
   * when set via a `Makefile`, the string we came up with (semver of most recent tag / v0.0.0 if none, appended by `-<hash>` if `HEAD` is ahead of most recent tag)
   * when not set via a `Makefile`, I guess `devel`
1. We do not want to have to set version multiple times via `-X`
1. We do not want to reimplement all of `Run()`; we are willing to reimplement `--version` flag

Given all of the above, we could:

* go back to our own `metal.VERSION` var
* default that to `"devel"`
* use that in the UserAgent

But then how do we get the `Version` line in the logs, which depends on `version.gitVersion` (private var)?

displague: Confirmed that the logged version is only settable via the version.Get / k8s packages:
https://github.com/kubernetes/cloud-provider/blob/master/app/controllermanager.go#L141-L144

I think we're stuck with the `-X` flags for this.

> Given all of the above, we could:

`--version` => "Cloud Provider Equinix Metal v3.3.0" would be nice, but we could address that as a separate issue since the original is about the logged version of the provider and overriding options may be tricky.

deitch: So what are you thinking? Leave this one as is for now?

I still would like to solve the `--version` problem, but this may be good enough for now.

displague (approved):

Makefile
displague: Noting that we discussed this elsewhere and discovered:
`-X "k8s.io/component-base/version/verflag.programName=Cloud Provider Equinix Metal"'`:

```
go run -ldflags '-extldflags "-static" -X "github.com/equinix/cloud-provider-equinix-metal/metal.VERSION=5.6.7" -X "k8s.io/component-base/version.gitVersion=v4.3.2"  -X "k8s.io/component-base/version/verflag.programName=Cloud Provider Equinix Metal"' ./... --version
Cloud Provider Equinix Metal v4.3.2
```

deitch: And thanks for putting it in with #223 !

PR #222, opened 2021-12-16
use labels and env vars to set EIP metro or facility, else error

### LINKED ISSUE #183: Metadata usage may be problematic
The Equinix Metal metadata service is currently consumed under conditions where the Facility was not specified via configuration.

The metadata is fetched from any node in the cluster. There are some problems with this:
- using metadata to determine the facility of one node in the cluster, does not guarantee that other nodes in the cluster are located in the same facility. The default portal behavior is to provision nodes in metros, which could result in nodes landing in multiple facilities.
- in a heterogenous cloud setup, the ccm could be running on nodes not in EM (while managing EM nodes). the metadata service would not be available.

One example of metadata use is: https://github.com/equinix/cloud-provider-equinix-metal/blob/b686e3fd36bfa8bf2aefd890e2bef168d8b8b620/main.go#L136-L144

deitch:
Fixes #183 

Completely eliminates the usage of EQXM metadata in CPEM. Instead, follows similar logic suggested in [this comment](https://github.com/equinix/cloud-provider-equinix-metal/issues/183#issuecomment-903797889) by @detiber:

1. If `Service.Spec.LoadBalancerIP` is set, we use it, nothing more to do; if not...
1. If `Service.Metadata.Annotations["topology.kubernetes.io/region"]` is set, get an EIP for the given metro; if not...
1. If `Service.Metadata.Annotations["topology.kubernetes.io/zone"]` is set, get an EIP for the given facility; if not...
1. If the env var / cli flag for metro is set on CCM, get an EIP for the given metro; if not...
1. If the env var / cli flag for facility is set on CCM, get an EIP for the given facility; if not...
1. Get no EIP

Does not address the "global EIP" option; we can add it later.

Updated README as well, of course


README.md
```diff
@@ -283,11 +290,18 @@ spec:
 
 CCM will detect that `loadBalancerIP` already was set and not try to create a new Equinix Metal Elastic IP.
 
-If the `Service.Spec.LoadBalancerIP` was *not* set, then CCM will use the Equinix Metal API to request a new, facility-specific
-Elastic IP and set it to `Service.Spec.LoadBalancerIP`.
+##### Equinix EIP
```
displague, line 293: We should link to this section in step 6 of the Metallb loadbalancing details and step 3 of the kube-vip loadbalancing details, both of which say ~ "a new EIP will be created". If we link to this section it will clear up how that EIP will be created.

metal/loadbalancers.go
```diff
@@ -340,6 +341,8 @@ func (l *loadBalancers) reconcileServices(ctx context.Context, svcs []*v1.Servic
 func (l *loadBalancers) addService(ctx context.Context, svc *v1.Service, ips []packngo.IPAddressReservation) error {
 	svcName := serviceRep(svc)
 	svcTag := serviceTag(svc)
+	svcRegion := serviceRegion(svc)
```
displague, line 344: Should this be nodeRegion and nodeZone?
https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion
Topology annotations are not defined on the Service, afaict. The CCM applies these annotations at the Node level, so the only way a Service would possess them is if the user assigned them manually.

displague: Without diving through the code, I'm not sure where / how you will get access to the relevant node resource from this function. This also has me wondering what would happen if two nodes had different topology settings.

I think we can dismiss varying `metro` fields for now (although, this could be possible via https://metal.equinix.com/developers/docs/networking/backend-transfer/).

It is also possible that facility values could be set without metro (strictly by the logic provided) and in that case, what would we expect or want to happen if a service needed an EIP and it was forced to create either an EIP from one facility or another based on node annotations.  Fortunately, I believe that all facility created EIPs are metro portable, but that wasn't the case a few months ago.

README.md
deitch: That is a really good point, helps close the loop. I will add that.

metal/loadbalancers.go
deitch: > Should this be nodeRegion and nodeZone?

No, it is specifically Service (and that was @detiber's great idea: use the official labels but apply them to services). 

What we are telling the user is: look, we have no way of knowing what nodes will be responding to the external requests, if you will have dedicated ingress nodes or all nodes, if you will split across data centers. And we really don't want to know. We don't know what will happen and where the nodes will be, so we trust that you know what you are doing. So we will leave it in your control as follows:

1. If you want the load balancer EIP to be in a specific metro or region, set the official label _on the `Service`_ and we will respect that (metro/region > facility/zone).
2. If you haven't defined it on the `Service`, then we will look for a fallback metro/region or facility/zone for the entire cluster, as set on the CCM when you started it. 
3. If you haven't defined it on the `Service` or on the CCM as a fallback, we cannot assign an EIP, so we won't.

Does that make sense?

README.md
deitch: Done. Thanks again for the good suggestion!

metal/loadbalancers.go
displague: I'm reluctant to apply a well-known annotation in a way that is not discussed in the well-known documentation.  (Nodes and PersistentVolumes usage is explicitly defined).

While digging into this further, I stumbled upon the following which may offer a solution (I haven't fully groked it):

https://v1-20.docs.kubernetes.io/docs/concepts/services-networking/service-topology/

deitch: > I'm reluctant to apply a well-known annotation in a way that is not discussed in the well-known documentation. 

I understand that reluctance. On the other hand, we have a very real problem - need to determine where to create an EIP, currently defaulting to whatever facility/metro the node running CPEM is in, which is unacceptable - which needs a real solution. I particularly like using @detiber's solution that uses labels (which can be arbitrary), adopts official labels that are meant to represent zones and regions, and fit within how those often are used. This doesn't go _against_ their usage, it does _expand_ their usage.

The alternative is to just use arbitrary labels (or annotations), which would be much worse, and far more likely to diverge from whatever is adopted in future standards (if anything).

> I stumbled upon the following which may offer a solution (I haven't fully groked it):

Not really AFAICT. It is meant to control what happens to traffic once it is on-cluster. So if a Service receives traffic on a particular node - from external ingress, nodeport, another pod, whatever - it could have sent it to any endpoint backing the service. That might mean that it would send it halfway across the world when one of those endpoints might have been on that exact same node.

The service topology allows you to create preferential order, e.g., "when a service processes a connection request, only send it to an endpoint on the same node" or "prefer to send it to an endpoint on the same node, accept an endpoint in the same region."

It doesn't really offer us much in terms of the ability to control in which region an EIP should be created.

deitch: > I understand that reluctance. On the other hand, we have a very real problem

If we can come up with a better solution, I would be all for it. AFAIK, the Kubernetes community has no standard around declaring where the EIP should be created, largely because it just uses local zonal load balancers.

displague: I've finally read through the service topology feature and you summed it up nicely. While this will be helpful for keeping packets in-region for multi-region (facility) deployments, this doesn't help us to determine which EIP type / location should be used for a service.  One thing that I would like to see is for us to put the switch-short-uuid in a EM specific node annotation so that service topology can prefer the same ToR switch. This won't be possible until service topology extends the valid list of annotations beyond the well-known four that are supported today.

My concern with this PR is that we are repurposing a well-known annotation. When a user sees this, they will have preconceived ideas about what it means and how it is applied. At first guess, I might assume the annotation was misplaced because the documentation does not say that these topology annotations are valid for services.  In the future, perhaps they will be.  At that time, we'll have to change the annotation that we use.

Can we come up with a CPEM specific annotation for this purpose?
- `metal.equinix.com/eip-facility`
- `metal.equinix.com/eip-metro`
- `metal.equinix.com/eip-reservation-id`
- ...

We wouldn't have to cross jargons. If this is a EM API flag, we may as well use EM API jargon.
We already offer a set of custom annotations for BGP settings. This also seems to be what AWS, GCP, etc have done:
- https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.3/guide/service/annotations/#legacy-cloud-provider
- https://github.com/kubernetes/legacy-cloud-providers/blob/master/gce/gce_annotations.go#L34-L84

In determining how the EIP locality should be determined, I understand that we want to avoid using local metadata - in part because the locality stored in the metadata of the node running the code may not match the locality of the service that needs an EIP.  The node running CPEM may not be an EM machine at all.

However, I think in most cases, we can expect that a node is running in the metro where the service EIP is intended. I wonder if step 6 ("No EIP") should be moved to step 7 with one last attempt made: "EIP will be requested in the metro of the node running CPEM, if the node is an EM node". This would allow for CPEM to run with less configuration (environment variables, flags) assuming we don't need this flag elsewhere.  If we add this behavior, it would be reasonable to introduce a flag that disables this self-discovery behavior.

deitch: I think we are on the same page here, just need agree on which of these we use:

* the official Kubernetes labels used to represent region/zone, but applied somewhere they are not today (services)
* EQXM specific annotations applied to services

Also note that one is labels, one is annotations.

I have a leaning towards the first, primarily because I dislike the proliferation of annotations everywhere. However, I am not set against it, and I do understand your concerns about confusion and about it applying to APIs. 

Let's ask @detiber to weigh in on this, as it was his original idea. If you both prefer the second option, I will change this to match that. If @detiber prefers the official labels applied to Services (the way this PR is), then let's go that way - going for the majority of people on this one.

detiber: Considering @displague's concerns, I'm good with using metal specific annotations and we can always re-evaluate if it makes sense to consolidate with upstream topology usage down the line if it makes sense.

deitch: Done.

deitch:
Done. It now determines where to place the EIP as follows:

1. Global facility set by env var; else
2. Global metro set by env var; else
3. Service-specific facility annotation; else
4. Service-specific metro annotation; else
5. Fail

The annotation follows the pattern of all other annotations we use: `metal.equinix.com/*`. In this case, `metal.equinix.com/eip-metro` and `metal.equinix.com/eip-facility`. Like all our annotations, that annotation is a default, and is configurable.

Ready for another full review @displague and @detiber

deitch:
> We have not updated the Helm chart in this PR (https://github.com/equinix/cloud-provider-equinix-metal/tree/master/deploy/chart). Do you think the new location options should be required?

Yes, but it can afford a separate PR. We should check all of the deployment stuff.

> I worry about deploying this change in a release with "Step 5 Fail" because anyone upgrading to this image, via latest or a minor release bump and takes advantage of EIP features will get errors upon update until the new environment variables / command line options are added.

Agreed, but that is what semver is for. Someone who uses `latest` does so at their own risk.

> Perhaps we can follow up this PR with one that adds a default behavior to check EM metadata (or API data). This could then be disabled in use-cases where it does not make sense. Upgrades would not break and the Helm chart would not need to require a location parameter. What do you think?

Location parameter only is required if you do not have it on the service annotations. I am comfortable with it being this way

metal/loadbalancers.go
```diff
@@ -388,6 +393,12 @@ func (l *loadBalancers) addService(ctx context.Context, svc *v1.Service, ips []p
 			// if we did not find an IP reserved, create a request
 			klog.V(2).Infof("no IP assignment found for %s, requesting", svcName)
 			// create a request
+			// our logic as to where to create the IP:
+			// 1. if metro is set globally, use it; else
+			// 2. if facility is set globally, use it; else
+			// 3. if Service.Metadata.Labels["topology.kubernetes.io/region"] is set, use it; else
+			// 4. if Service.Metadata.Labels["topology.kubernetes.io/zone"] is set, use it; else
```
displague, line 400: A user is able to offer conflicting annotations vs other settings.
The rules spelled out how this will be processed so I can't complain.

When both Region and Zone are set, especially when the zone resides in the region, I think it's reasonable to assume the more granular, Zone, will be used.  This is not the case today.  Based on how EM EIPs work, I think users supplying both settings will be better served with the Regional EIP (Metro based).

metal/loadbalancers.go
```diff
@@ -388,6 +393,12 @@ func (l *loadBalancers) addService(ctx context.Context, svc *v1.Service, ips []p
 			// if we did not find an IP reserved, create a request
 			klog.V(2).Infof("no IP assignment found for %s, requesting", svcName)
 			// create a request
+			// our logic as to where to create the IP:
+			// 1. if metro is set globally, use it; else
```
displague, line 397: Perhaps the more granular setting, the service annotation, should be consulted first with the "default" coming from the global settings?

displague: Administrators are more likely to have access to the CPEM runtime parameters.
Users are more likely to have access to the service annotation.

displague: If the annotation values permitted need to be restricted, we could introduce another global setting to define the permitted values.

deitch: You mean to use the env var / CLI flag as a _fallback_ rather than _override_? I can see the use case, but that is not how it has been used in the past. In any case, the only ones ever likely to use env var on the actual CPEM will be administrators. If they set it, they _really_ mean to override.

I don't mind considering having another env var for fallback, but this one historically has been the explicit override; I would like to leave it as is.

deitch: The flags/env vars really are just meant as controlling overrides for development time. They don't have a good production purpose.

displague (dismissed):

deitch:
Rebased, letting CI run

deitch:
Needs a new approval after rebase @displague

displague (approved):

PR #224, opened 2022-01-16
use standard API LoadBalancers

deitch:
This is a deep change in how CCM operates.

### Existing

The CCM has listeners for changes to services and nodes, as well as a timer loop. Every time a node or service is added or removed, or a node is added or removed, it calls the `serviceReconciler` or `nodeReconciler` or each sub-service - bgp, loadbalancers, devices, facilities, control-plane-EIP. Some of these return `nil`, indicating "nothing to do".

The main ones that handle it are bgp and load balancer. 

* services: looks at the added service, filters for `type=LoadBalancer`, gets an EIP from EQXM if needed, and then calls the specific implementation of load balancer to add or remove it
* nodes: looks at the node, filters out for specific labels if provided, and then ensures bgp is enabled on the node, adds annotations, and tells the specific LB implementation about it

The problems with the existing design are as follows:

* does not follow the [official cloud provider Interface](https://pkg.go.dev/k8s.io/cloud-provider#Interface), which expects to manage a standard set of [LoadBalancer calls](https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer); this makes understanding what this does harder
* because of the previous one, we have to do a lot of handling of state, sync, etc. that shouldn't be the CCM's problem
* spreads the work to different areas
* is more complex to handle and reason about

### Proposed

Use the provided standard [LoadBalancer API](https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer).

In doing so, we eliminate almost all of the watcher and timer loop calls, save the ones to the control plane EIP, that will have to be figured out separately. 

This does, however, have a different mindset. In the ones we have done, nodes are added/removed and services added/removed, but no connection between them. In the official k8s mindset as it relates to LBs, nodes are simply endpoints for an LB. Thus, the idea of "add a node" makes no sense. Instead, it says, "add a Load Balancer and here are the nodes for it", "update a load balancer with these nodes", and "remove a load balancer and any of its attendant nodes".

To do this, we had to changes what we send to the specific LB implementation and how. It actually simplified the interface.

## Challenges

Most of these changes affect only the CCM itself or one specific LB implementation, i.e. metallb. Others (mainly kube-vip) do not rely on those and just use the annotations as they appear.

For metallb, we need some way of tracking which nodes are announcing for which services, i.e. which nodes (announced) "back" which LB (EIP). This is not inherent anywhere in the metallb config file. We cannot just store it in memory, as that is not persistent. The metallb config file has no comment or description fields, for either address pool or peer listing, so that doesn't help. 

I did something rather ugly by overloading the `NodeSelector.MatchLabel` to have one that indicates what service it is.

```yaml
peers:
- peer-address: 10.0.0.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
      kubernetes.io/hostname: my-node
      nomatch.metal.equinix.com/service-name: serviceA
```

I will be the first to admit that is _ugly_, and wrong. But I couldn't come up with a cleaner way to start it. At least it should provoke the conversation.

The other downside is that we now need one peer entry for each service. Actually, we need 2, one for each upstream ToR router. If we have 20 services across 5 nodes, instead of 10 entries, we now need 200 (2 per service per node).

So I really don't want this to get merged in as is. But I would like some better ideas as to how to manage it.

### DIFF
--- a/metal/bgp.go
+++ b/metal/bgp.go
@@ -1,57 +1,31 @@
 package metal
 
 import (
-	"context"
-	"encoding/base64"
-	"encoding/json"
 	"fmt"
-	"sort"
-	"strconv"
 	"strings"
 
 	"github.com/packethost/packngo"
 	"github.com/pkg/errors"
 
-	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	k8stypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
 
 type bgp struct {
-	project            string
-	client             *packngo.Client
-	k8sclient          kubernetes.Interface
-	localASN           int
-	bgpPass            string
-	annotationLocalASN string
-	annotationPeerASN  string
-	annotationPeerIP   string
-	annotationSrcIP    string
-	annotationBgpPass  string
-	nodeSelector       labels.Selector
+	project   string
+	client    *packngo.Client
+	k8sclient kubernetes.Interface
+	localASN  int
+	bgpPass   string
 }
 
-func newBGP(client *packngo.Client, project string, localASN int, bgpPass string, annotationLocalASN, annotationPeerASN, annotationPeerIP, annotationSrcIP, annotationBgpPass string, nodeSelector string) *bgp {
-
-	selector := labels.Everything()
-	if nodeSelector != "" {
-		selector, _ = labels.Parse(nodeSelector)
-	}
+func newBGP(client *packngo.Client, project string, localASN int, bgpPass string) *bgp {
 
 	return &bgp{
-		project:            project,
-		client:             client,
-		localASN:           localASN,
-		bgpPass:            bgpPass,
-		annotationLocalASN: annotationLocalASN,
-		annotationPeerASN:  annotationPeerASN,
-		annotationPeerIP:   annotationPeerIP,
-		annotationSrcIP:    annotationSrcIP,
-		annotationBgpPass:  annotationBgpPass,
-		nodeSelector:       selector,
+		project:  project,
+		client:   client,
+		localASN: localASN,
+		bgpPass:  bgpPass,
 	}
 }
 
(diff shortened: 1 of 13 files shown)


deitch:
To clarify this one a bit further, and hopefully get some good suggestions on it. 

Prior to this possible change, the nodes that announced and the services were decoupled. You can see that in the API, but mainly in the metallb config. For example, with 2 services and 2 nodes:

```yaml
peers:
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
address-pools:
- name: serviceA
   protocol: layer2
   addresses:
   - 100.100.100.10
- name: serviceB
   protocol: layer2
   addresses:
   - 100.100.100.11
```

However, the native LoadBalancer API from cloud-provider always sends nodes as part of the service, because it thinks in terms of load-balancers (as it should), which are "here are the nodes for the load balancer for this service".

If we hew to that, we need to track which nodes are there for which LBs (represented as IPs). There could be complete, partial, or zero overlap.

The (ugly) suggestion I put in this PR would lead to:

```yaml
peers:
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-name: serviceA
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-name: serviceA
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-name: serviceB
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-name: serviceB
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceA
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceA
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceB
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceB
address-pools:
- name: serviceA
   protocol: layer2
   addresses:
   - 100.100.100.10
- name: serviceB
   protocol: layer2
   addresses:
   - 100.100.100.11
```

That does not exactly scale. When we have 5 nodes, 2 peers each, 20 services, that will be 5*2*20 = 200 entries! 

It also abuses the `match-labels` field in an ugly way. Unfortunately, there is no way in the metallb configmap to record arbitrary data, no annotations or description field, neither for the peers not for the address-pool.

Another possibility is to stick with the overloaded `match-labels`, but to have multiple services (leaving `address-pools` out for simplicity):

```yaml
peers:
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-names: serviceA,serviceB
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
       nomatch.metal.equinix.com/service-name: serviceA,serviceB
- peer-address: 169.254.255.1
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceA,serviceB
- peer-address: 169.254.255.2
  peer-asn: 64501
  my-asn: 64500
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceA,serviceB
```

That would require additional complex logic in terms of determining if a host-peer entry already exists, but not _too_ bad. I didn't want too put in the effort to make that real until I knew we were comfortable with the `match-labels`-overload approach.

Would like feedback from @displague , @cprivite , @ocobleseqx , @thebsdbox and especially @detiber (who really knows the ins and outs of where load balances are now and will be going).

displague:
We've talked about some of these changes in the past. I’ll have to take a closer look, but at first glance it should be great to remove the custom watchers/informers/caches.  

Keeping the same UX and configuration file while getting this benefit is also excellent. 

It seems the only user-facing change is that the CCM will inject and parse a metallb config parameter. Metallb will be ignorant and unaffected by this extra parameter.

I haven't considered alternative approaches yet, but would the given approach break the peer node matching?

Given https://github.com/metallb/metallb/blob/main/manifests/example-config.yaml#L34-L40, do we need two separate selectors? One to act as arbitrary metadata and another to get the desired node matching behavior?

```
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node1
  - match-labels:
       nomatch.metal.equinix.com/service-name: serviceA
 ```

**Comment by deitch** [comment]
Yes, @displague you are 100% right; I got them backwards. The language is:

> A node
> matches if at least one of the node selectors matches. Within
> one selector, a node matches if all the matchers are
> satisfied.

So instead of:

```yaml
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
       nomatch.metal.equinix.com/service-name: serviceA,serviceB
```

it should be:

```yaml
  node-selectors:
  - match-labels:
       kubernetes.io/hostname: node2
  - match-labels:
       nomatch.metal.equinix.com/service-name: serviceA,serviceB
```

Either way, are we comfortable with this?

deitch:
Fixed the fmt-check issue, and changed it to do the correct node-selector as you caught (thanks for that one @displague). 

Started the logic for handling the peers correctly, will make progress, but want concurrence that we are ok with this path.

deitch:
I mostly got the logic in place, tests require updating and expansion, so CI will fail. But worth taking a look.

deitch:
Well, then, tests are in. I need to test against live clusters, but let's get some final concurrence on the design first.

deitch:
Phew. Took some detailed work on testing and fixing some odd issues, but this works solidly now. 

This is ready to go in @displague @detiber . Once approved, I will merge. After that, I can move all of the watcher stuff into EIP control plane manager, since that will be the only place that needs it. It will make everything much simpler to read and understand.

thebsdbox (approved):
Given all the other comments have been largely addressed, I think we should be OK to move forward with this. We have time to shake out any unknowns before cutting an official release.

displague (approved):

PR #228, opened 2022-02-02
do not log EM device objects which may contain sensitive fields

displague:
Devices contain "RootPassword" and other fields that may contain sensitive fields. Keep this data out of the logs.


deitch (approved):
It is part of our security posture; when everyone sees that their passwords are logged, they will stop using them! 😆 

Thanks for doing this.

PR #230, opened 2022-03-01
[WIP] Refactor EIP handling

detiber:
Work in progress to rebase the external EIP handling changes I've been working on to the latest codebase.


Dockerfile_builder
```diff
@@ -1,11 +0,0 @@
-FROM golang:1.10.1-alpine3.7
```
detiber, line 1: As far as I could tell, this file was unused

metal/bgp.go
```diff
@@ -19,7 +18,7 @@ type bgp struct {
 	bgpPass   string
 }
 
-func newBGP(client *packngo.Client, k8sclient kubernetes.Interface, stop <-chan struct{}, project string, localASN int, bgpPass string) (*bgp, error) {
```
detiber, line 22: stop was unused

metal/cloud.go
```diff
@@ -31,8 +31,6 @@ type cloud struct {
 	config                      Config
 	instances                   cloudprovider.InstancesV2
 	loadBalancer                cloudprovider.LoadBalancer
-	metro                       string
-	facility                    string
```
detiber, line 35: These were unused

metal/config.go
```diff
@@ -114,10 +113,6 @@ func getMetalConfig(providerConfig io.Reader) (Config, error) {
 	if loadBalancerSetting != "" {
 		config.LoadBalancerSetting = loadBalancerSetting
 	}
-	// and set for default
-	if config.LoadBalancerSetting == "" {
-		config.LoadBalancerSetting = defaultLoadBalancerConfigMap
-	}
```
detiber, line 120: As far as I can tell this didn't actually work as a default, since `metallb-system:config` does not parse as a URL.

metal/devices.go
```diff
@@ -21,140 +21,31 @@ type instances struct {
 }
 
 var (
-	_ cloudprovider.Instances   = (*instances)(nil)
```
detiber, line 24: Removed the old Instances interface support.

metal/ipreservations.go
```diff
@@ -42,38 +42,3 @@ ips:
 	}
 	return ret
 }
-
```
detiber, line 45: These were unused

metal/loadbalancers.go
```diff
@@ -48,7 +48,7 @@ type loadBalancers struct {
 	nodeSelector          labels.Selector
 }
 
-func newLoadBalancers(client *packngo.Client, k8sclient kubernetes.Interface, stop <-chan struct{}, projectID, metro, facility, config string, localASN int, bgpPass, annotationNetwork, annotationLocalASN, annotationPeerASN, annotationPeerIP, annotationSrcIP, annotationBgpPass, eipMetroAnnotation, eipFacilityAnnotation, nodeSelector string) (*loadBalancers, error) {
+func newLoadBalancers(client *packngo.Client, k8sclient kubernetes.Interface, projectID, metro, facility, config string, localASN int, bgpPass, annotationNetwork, annotationLocalASN, annotationPeerASN, annotationPeerIP, annotationSrcIP, annotationBgpPass, eipMetroAnnotation, eipFacilityAnnotation, nodeSelector string) (*loadBalancers, error) {
```
detiber, line 51: stop was unused here

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +74,156 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
+						klog.Errorf("failed to sync service from default/kubernetes to %s/%s: %v", externalServiceNamespace, externalServiceName, err)
+						return
+					}
+				},
+			},
+		},
+	)
+
+	sharedInformer.Start(stop)
+	sharedInformer.WaitForCacheSync(stop)
```
detiber, line 224: This should have the same result as the custom code we had in `watchers.go`

Dockerfile_builder
deitch: Oh yes, this is pretty old. I don't know how we didn't even notice it still was around.

metal/bgp.go
deitch: Yes, good catch. After the last big PR, it no longer is needed

metal/cloud.go
deitch: Yes. Same as the last. These should have gone with the last big PR.

metal/config.go
deitch: Again, good catch. It didn't break anything, as the part that processes the URL only sets the configmap if it is a valid URL.

detiber:
Rebased now that #231 has been merged, but this PR also includes a fix from #232 as well.

deitch:
The merger of #232 should make this even smaller post-rebase?

detiber:
@deitch rebased on latest.

deitch:
Also, it doesn't look like it checks the EIP address, only the node's?

deitch:
> I was just running through another test of https://github.com/detiber/cluster-api-provider-packet/pull/65 that adds kube-vip support to the cluster-api provider and I noticed that the Equinix Metal CCM is assigning the EIP to one of the control plane nodes. However, I don't the think the CCM should be doing this.

It only does this if you pass it the right tag for the EIP. If you do not, CCM ignores it. That is the flag that says, "you should control the EIP" or "you should ignore it".

if you are using kibe-vip in its mode of managing the control plane EIP, then the CCM should not have the tag set.

DavidSpek:
@deitch Thanks for the explanation, I just found that as well and updated my PR to reflect the change.

detiber:
> @detiber I get most of it - looks good - except for the core. The existing one depends on `reconcileNodes()` being called once in a while (e.g. 30 seconds), and then checking the health of the EIP. If it failed, it then went on to check the control plane node IPs and move the EIP around, as needed.
> 
> In this proposed one, it appears to change that. It only checks the health in `doHealthCheck()`, which only is called by `UpdateFunc:`, and it only checks if the node on which `UpdateFunc` was called actually currently has the EIP assigned.
> 
> How does this regularly check the status of the EIP and update it if necessary? I get that it would work when `UpdateFunc` is called, but why would it update it with any regularity?

Because the resync interval on the informer is set, every time a resync happens `UpdateFunc` is called for each of the Nodes, so that keeps us doing a health check at most roughly every minute, which is approximately the same as the old timer loop happening every 60 seconds.

detiber:
> Also, it doesn't look like it checks the EIP address, only the node's?

Yeah, that was the bit that I suspected to get some push back on...  In the case of Cluster API, we are configuring the EIP as a local loopback address: https://github.com/kubernetes-sigs/cluster-api-provider-packet/blob/main/templates/cluster-template.yaml#L30-L36, this was needed to work around how `kubeadm` expects the address to be resolvable when bootstrapping (since the external service/endpoints would not be configured yet, since CPEM is not yet deployed).

This resulted in issues during upgrade, where the local etcd instance is removed from the cluster prior to the machine being deleted, which would result in the local apiserver being unresponsive. Not necessarily an issue if CPEM is running on the machine being deleted, but becomes an issue when it's one of the control plane nodes that isn't running CPEM that is also currently assigned the EIP. CPEM is attempting to health check the EIP, which resolves to localhost and passes, so doesn't attempt to re-assign the EIP.

deitch:
> Because the resync interval on the informer is set, every time a resync happens UpdateFunc is called for each of the Nodes, so that keeps us doing a health check at most roughly every minute, which is approximately the same as the old timer loop happening every 60 seconds.

Ah, I didn't realize that. So basically every update-interval for resync it automatically will call the health check? Then that makes sense.

deitch:
> Yeah, that was the bit that I suspected to get some push back on... In the case of Cluster API, we are configuring the EIP as a local loopback address: https://github.com/kubernetes-sigs/cluster-api-provider-packet/blob/main/templates/cluster-template.yaml#L30-L36, this was needed to work around how kubeadm expects the address to be resolvable when bootstrapping (since the external service/endpoints would not be configured yet, since CPEM is not yet deployed).

Oh yes, I have suffered through that. I do wish kubeadm was a bit more configurable in that regards.

> CPEM is attempting to health check the EIP, which resolves to localhost and passes, so doesn't attempt to re-assign the EIP.

So the reason you did it is so that we won't have a situation where the health check in CPEM _thinks_ everything is healthy - because it is checking the EIP, which is on local loopback, but really it isn't OK when coming from the Internet?

detiber:
> > Yeah, that was the bit that I suspected to get some push back on... In the case of Cluster API, we are configuring the EIP as a local loopback address: https://github.com/kubernetes-sigs/cluster-api-provider-packet/blob/main/templates/cluster-template.yaml#L30-L36, this was needed to work around how kubeadm expects the address to be resolvable when bootstrapping (since the external service/endpoints would not be configured yet, since CPEM is not yet deployed).
> 
> Oh yes, I have suffered through that. I do wish kubeadm was a bit more configurable in that regards.
> 
> > CPEM is attempting to health check the EIP, which resolves to localhost and passes, so doesn't attempt to re-assign the EIP.
> 
> So the reason you did it is so that we won't have a situation where the health check in CPEM _thinks_ everything is healthy - because it is checking the EIP, which is on local loopback, but really it isn't OK when coming from the Internet?

Exactly

detiber:
@deitch Do you think it would make sense to make using the EIP URL or not a configurable option?

I'm not sure how others may be using CPEM outside of Cluster API to know if it's a broader issue or just a localized one.

deitch:
I don't understand.

detiber:
@deitch rather than just avoiding the use of the EIP automatically (like this PR is doing) for health checks, or would it make sense to make that behavior opt-in, with the default continuing to use the EIP url for health checks?

deitch:
> @deitch rather than just avoiding the use of the EIP automatically (like this PR is doing) for health checks, or would it make sense to make that behavior opt-in, with the default continuing to use the EIP url for health checks?

Now that I get it, yes, I think so. My concern is lack of consistent EIP behaviour by CPEM, but that can be acceptable I think.

detiber:
> Would it make sense to split the conversation? This PR is changing two things simultaneously: how we invoke health checks; what health checks test.
> 
> Does this PR have any value doing just the first, and then having a separate one focused just on the second? Or must it go together?

Both are really needed in order to unblock the CAPI v1 support for CAPP, so I'm not sure there is much value in breaking it up.

That said, more than happy to break it up if that is your preference.

detiber:
This should be ready to go.

deitch:
Reviewing now. @detiber can you add the new option env var to the README?

metal/eip_controlplane_reconciliation.go
```diff
@@ -210,14 +233,9 @@ func (m *controlPlaneEndpointManager) reassign(ctx context.Context, nodes []*v1.
 		return errors.New("control plane node apiserver port not yet determined, cannot reassign, will try again on next loop")
 	}
 	for _, node := range nodes {
-		md, err := m.instances.InstanceMetadata(ctx, node)
-		if err != nil {
-			return err
-		}
-
```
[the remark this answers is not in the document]
deitch: And a lot more efficient, too.

metal/eip_controlplane_reconciliation.go
deitch: I am glad to be rid of that custom code.

metal/config.go
```diff
@@ -233,6 +235,15 @@ func getMetalConfig(providerConfig io.Reader) (Config, error) {
...
 
+	config.EIPHealthCheckUseHostIP = rawConfig.EIPHealthCheckUseHostIP
+	if v := os.Getenv(envVarEIPHealthCheckUseHostIP); v != "" {
+		useHostIP, err := strconv.ParseBool(v)
+		if err != nil {
+			return config, fmt.Errorf("env var %s must be a boolean, was %s: %w", envVarEIPHealthCheckUseHostIP, v, err)
+		}
+		config.EIPHealthCheckUseHostIP = useHostIP
+	}
+
```
deitch, line 246: We should add these to the README.

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +75,157 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
-		instances:     i,
 		ipResSvr:      ipResSvr,
 		deviceIPSrv:   deviceIPSrv,
 		apiServerPort: apiServerPort,
 		k8sclient:     k8sclient,
-		stop:          stop,
+		useHostIP:     useHostIP,
 	}
-	klog.V(2).Info("controlPlaneEndpointManager.init(): enabling BGP on project")
+
```
deitch, line 84: At later points, when processing each node, we check if `m.eipTag == ""` in order to know to do nothing. Should we not do it here as well? Why set this all up if `eipTag == ""`?

metal/eip_controlplane_reconciliation.go
```diff
@@ -267,106 +287,273 @@ func (m *controlPlaneEndpointManager) reassign(ctx context.Context, nodes []*v1.
...
+	for _, label := range controlPlaneLabels {
+		if metav1.HasLabel(node.ObjectMeta, label) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (m *controlPlaneEndpointManager) nodeFilter(n *v1.Node) bool {
```
deitch, line 300: Can we add a comment here what this filter does? Specifically that it:

* applies only to control plane nodes
* only if the `eipTag` is set
* only if the `apiServerPort` has been discovered (i.e. might be too early)

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +74,156 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
-			}
+	sharedInformer := informers.NewSharedInformerFactory(k8sclient, checkLoopTimerSeconds*time.Second)
+
+	// register to capture all new nodes
+	sharedInformer.Core().V1().Nodes().Informer().AddEventHandler(
+		cache.FilteringResourceEventHandler{
+			FilterFunc: func(obj interface{}) bool {
+				n, _ := obj.(*v1.Node)
+				return m.nodeFilter(n)
+			},
```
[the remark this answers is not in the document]
deitch: Given that the `apiServerPort` might not be set the first time a node is added, will this work the next time? Is the `FilterFunc` called on every sync?

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +74,156 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
+			Handler: cache.ResourceEventHandlerFuncs{
+				AddFunc: func(obj interface{}) {
```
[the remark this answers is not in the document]
deitch: `AddFunc` is the only place that we call `m.tryDefaultAPIServerPorts()`. Does that need to be called?

metal/eip_controlplane_reconciliation.go
```diff
@@ -267,106 +287,273 @@ func (m *controlPlaneEndpointManager) reassign(ctx context.Context, nodes []*v1.
...
+
+func (m *controlPlaneEndpointManager) syncEndpoints(ctx context.Context, k8sEndpoints *v1.Endpoints) error {
+	m.endpointsMutex.Lock()
+	defer m.endpointsMutex.Unlock()
+
+	eps, err := m.k8sclient.CoreV1().Endpoints(externalServiceNamespace).Get(ctx, externalServiceName, metav1.GetOptions{})
+
+	switch {
+	case apierrors.IsNotFound(err):
+		externalEndpoints := &v1.Endpoints{
```
deitch, line 489: The code in this case and the `default` case is almost identical. Would it work if we did it:

```go
	var externalEndpoints *v1.Endpoints
	switch {
	case apierrors.IsNotFound(err):
		externalEndpoints = &v1.Endpoints{
			ObjectMeta: metav1.ObjectMeta{
				Name:      externalServiceName,
				Namespace: externalServiceNamespace,
			},
		}

	case err != nil:
		return fmt.Errorf("failed to get endpoint %s/%s: %w", externalServiceNamespace, externalServiceName, err)
	default:
		externalEndpoints = eps.DeepCopy()
	}

	externalEndpoints.Subsets = make([]v1.EndpointSubset, 0, len(k8sEndpoints.Subsets))

	for _, subset := range k8sEndpoints.Subsets {
		externalEndpoints.Subsets = append(externalEndpoints.Subsets, *subset.DeepCopy())
	}

	// TODO: replace with Apply
	_, err := m.k8sclient.CoreV1().Endpoints(externalServiceNamespace).Update(ctx, externalEndpoints, metav1.UpdateOptions{})
	if err != nil {
		return fmt.Errorf("failed to update endpoint %s/%s: %w", externalServiceNamespace, externalServiceName, err)
	}

	return nil
```

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +74,156 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
-	if nodesInformer != nil {
-		watchers = append(watchers, nodesInformer)
-	}
+	)
 
-	if err := startWatchers(ctx, watchers); err != nil {
-		return nil, fmt.Errorf("watchers initialization failed: %v", err)
-	}
-	return m, nil
+	sharedInformer.Core().V1().Endpoints().Informer().AddEventHandler(
```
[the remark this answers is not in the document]
deitch: So is the point of `syncEndpoints` to get the endpoints, and the point of `syncService` to get the service port?

deitch (changes requested):
This is rather nicely done, and I think simplifies a lot.

I mostly have some questions, README requests, and some small changes

metal/eip_controlplane_reconciliation.go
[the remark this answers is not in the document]
detiber: FilterFunc is called prior to calling any of the Handler functions. So it'll be called for any watch based events or resync events.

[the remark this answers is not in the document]
detiber: Since we were attempting to default these values in both the node filter and during Services reconciliation, I went ahead and deduplicated it to only be done during the Services reconciliation instead.

metal/eip_controlplane_reconciliation.go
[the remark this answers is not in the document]
detiber: I cleaned that up to only be done as part of the Services reconciliation now, so I went ahead and removed the AddFunc Handler.

metal/eip_controlplane_reconciliation.go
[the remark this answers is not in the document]
detiber: Pretty much, yes

metal/eip_controlplane_reconciliation.go
detiber: As you suggested, I moved the validation for eipTag to avoid configuring all the reconciliation logic if it's not set.

metal/eip_controlplane_reconciliation.go
detiber: I removed the function definition, moving the logic to where we are defining the EventHandler. I also added additional comments there to help clarify

metal/eip_controlplane_reconciliation.go
detiber: I went ahead and refactored things to use ServerSideApply to avoid the duplication.

detiber:
I still need to address the README updates, but did quite a bit of cleanup based on the other comments.

Working on validating that the Server Side Apply changes didn't break anything, though 😬

detiber:
Server side apply changes seem to be working good, will pick back up on the README changes tomorrow.

metal/eip_controlplane_reconciliation.go
```diff
@@ -267,163 +279,349 @@ func (m *controlPlaneEndpointManager) reassign(ctx context.Context, nodes []*v1.
...
+	}
+
+	if len(schedulableNodes) > 0 {
+		return schedulableNodes
+	}
+
+	return nodes
+}
+
+type nodeFitler func([]*v1.Node) []*v1.Node
```
deitch, line 369: I think you meant for this to be `nodeFilter` (instead of `nodeFitler`, `t` and `l` reversed). Same for where it is consumed in a two other places

metal/eip_controlplane_reconciliation.go
deitch: And I rather like serversideapply. :-)

deitch:
This all looks solid. Just needs the README updated and that `nodeFitler` fixed, and we should be good to go.

metal/eip_controlplane_reconciliation.go
detiber: 😬 oops, yeah. I have no idea what a fitler is...

detiber:
@deitch README is updated and we are no longer fitlering anything.

I also squashed down the commits and cleaned up the commit message.

metal/eip_controlplane_reconciliation.go
```diff
@@ -69,138 +80,144 @@ func newControlPlaneEndpointManager(k8sclient kubernetes.Interface, stop <-chan
...
+					klog.Errorf("control plane apiserver ports not provided or determined, skipping: %s", n.Name)
+					return false
+				}
+
+				// only reconcile control plane nodes
+				return isControlPlaneNode(n)
+			},
+			Handler: cache.ResourceEventHandlerFuncs{
+				UpdateFunc: func(old, obj interface{}) {
+					oldNode, _ := obj.(*v1.Node)
```
deitch, line 123: Should this be:

```go
oldNode, _ := old.(*v1.Node)
```

It looks like we are using `obj` for both `oldNode` and `newNode`.

deitch (changes requested):
This all looks good, other than the single bug I think I might have found.

Comment here when that is patched, and I can run another series of tests on it.

detiber:
@deitch sgtm, I've been testing this code pretty regularly from an EIP perspective with the automated testing I've been doing for CAPP

deitch:
Really nice on this one, thank you.

PR #231, opened 2022-03-02
General cleanup and updating dependencies

detiber:
- Update go version to v1.17
- Update alpine to v3.15
- Update packngo to v0.22.0
- Update k8s dependencies to v0.23.4
- Cleanup unused code, constants, and variables
- fix some lint related issues
- Update error wrapping to use go v1.13+ style error wrapping instead of github.com/pkg/errors


detiber:
These are the cleanup changes from #230

go.mod
```diff
@@ -1,21 +1,108 @@
...
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.21.2
-	k8s.io/apimachinery v0.21.2
-	k8s.io/client-go v0.21.2
-	k8s.io/cloud-provider v0.21.2
-	k8s.io/component-base v0.21.2
-	k8s.io/klog/v2 v2.8.0
+	k8s.io/api v0.23.4
+	k8s.io/apimachinery v0.23.4
```
deitch, line 14: Does this create any backwards-compatibility issues with older cluster versions? Or do we just say, "use this version x of CCM for version y of cluster"?

go.mod
```diff
@@ -1,21 +1,108 @@
...
+	k8s.io/client-go v0.23.4
+	k8s.io/cloud-provider v0.23.4
+	k8s.io/component-base v0.23.4
+	k8s.io/klog/v2 v2.30.0
+)
+
+require (
+	github.com/Azure/go-ansiterm v0.0.0-20210617225240-d185dfc1b5a1 // indirect
```
deitch, line 22: That's quite the growth in explicit requirements. Why is that?

deitch (approved):
This looks great. I had 2 questions, but I would not block on them.

go.mod
detiber: It shouldn't cause any issues, since client-go has a pretty good track record of backward/forward compatibility

go.mod
detiber: It's new behavior in go 1.17 related to how go builds up the dependency graphs for projects: https://go.dev/doc/go1.17#go-command

deitch:
Testing all was clean. Thanks for the cleanup @detiber

PR #232, opened 2022-03-02
Fix issue with services controller when load balancing should be disabled

detiber:
When trying to run with the load balancer disabled, currently the controller will panic, because we are testing if an interface is nil rather than if the value of the interface is nil).

Replacing the interface types in the cloud struct with the concrete types resolves the issue and as far as I can tell does not impact anything else.

### DIFF
--- a/metal/cloud.go
+++ b/metal/cloud.go
@@ -29,8 +29,8 @@ const (
 type cloud struct {
 	client                      *packngo.Client
 	config                      Config
-	instances                   cloudprovider.InstancesV2
-	loadBalancer                cloudprovider.LoadBalancer
+	instances                   *instances
+	loadBalancer                *loadBalancers
 	controlPlaneEndpointManager *controlPlaneEndpointManager
 	// holds our bgp service handler
 	bgp *bgp
--- a/metal/cloud_test.go
+++ b/metal/cloud_test.go
@@ -28,11 +28,6 @@ const (
 	validPlanName   = "Bill by the hour"
 )
 
-var (
-	validCloud *cloud
-	backend    *store.Memory
-)
-
 // mockControllerClientBuilder mock implementation of https://pkg.go.dev/k8s.io/cloud-provider#ControllerClientBuilder
 // so we can pass it to cloud.Initialize()
 type mockControllerClientBuilder struct {
@@ -60,13 +55,9 @@ func (a *apiServerError) Error(err error) {
 }
 
 // create a valid cloud with a client
-func testGetValidCloud(t *testing.T) (*cloud, *store.Memory) {
-	// if we already have it, it is fine
-	if validCloud != nil {
-		return validCloud, backend
-	}
+func testGetValidCloud(t *testing.T, LoadBalancerSetting string) (*cloud, *store.Memory) {
 	// mock endpoint so our client can handle it
-	backend = store.NewMemory()
+	backend := store.NewMemory()
 	fake := emServer.PacketServer{
 		Store: backend,
 		ErrorHandler: &apiServerError{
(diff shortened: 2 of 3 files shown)


metal/cloud.go
```diff
@@ -29,8 +29,8 @@ const (
 type cloud struct {
 	client                      *packngo.Client
 	config                      Config
-	instances                   cloudprovider.InstancesV2
-	loadBalancer                cloudprovider.LoadBalancer
+	instances                   *instances
+	loadBalancer                *loadBalancers
```
detiber, line 33: This fixes the issue where we were testing an interface type for nil to determine if the loadbalancer controller should be enabled or not.

metal/cloud_test.go
```diff
@@ -28,11 +28,6 @@ const (
 	validPlanName   = "Bill by the hour"
 )
 
-var (
-	validCloud *cloud
-	backend    *store.Memory
-)
```
detiber, line 34: Removing these variables was the easiest way that I could enable testing with different configurations instead of a common shared config.

metal/cloud_test.go
```diff
@@ -84,17 +75,66 @@ func testGetValidCloud(t *testing.T) (*cloud, *store.Memory) {
...
+		expectedSupported = false
+		expectedResponse  = response
+	)
+	if supported != expectedSupported {
+		t.Errorf("supported returned %v instead of expected %v", supported, expectedSupported)
+	}
+	if response != expectedResponse {
+		t.Errorf("value returned %v instead of expected %v", response, expectedResponse)
+	}
+}
```
detiber, line 101: This was previously returning true erroneously, now it properly returns false.

metal/cloud_test.go
```diff
@@ -84,17 +75,66 @@ func testGetValidCloud(t *testing.T) (*cloud, *store.Memory) {
...
+		expectedSupported = true
+		expectedResponse  = response
+	)
+	if supported != expectedSupported {
+		t.Errorf("supported returned %v instead of expected %v", supported, expectedSupported)
+	}
+	if response != expectedResponse {
+		t.Errorf("value returned %v instead of expected %v", response, expectedResponse)
+	}
 }
```
detiber, line 133: Skipping these because we don't have a k8s client plumbed through for the code paths to be able to run.

metal/cloud_test.go
```diff
@@ -84,17 +75,66 @@ func testGetValidCloud(t *testing.T) (*cloud, *store.Memory) {
...
 
-func TestLoadBalancer(t *testing.T) {
-	vc, _ := testGetValidCloud(t)
+func TestLoadBalancerKubeVIP(t *testing.T) {
+	t.Skip("Test needs a k8s client to work")
+	vc, _ := testGetValidCloud(t, "kube-vip://")
```
detiber, line 137: Skipping these because we don't have a k8s client plumbed through for the code paths to be able to run.

metal/cloud.go
deitch: We actually had pointers to the concrete types in one of the PRs, but switched them to the interface types at the suggestion of @displagueto makes it easier to pass things in for testing and keep it standard.

Where do we do that testing?

Also, is it not legitimate to test if an interface type is nil?

detiber: > Silly test, but look at https://go.dev/play/p/yC4BLAqvccn
> 
> I can legitimately test an interface type for nil.

Here is a better example of what I mean: https://go.dev/play/p/OkcY_Nvsc3H

Since we are setting the value with the returned value of a constructor, the type of the argument at runtime is not actually nil

detiber: > We actually had pointers to the concrete types in one of the PRs, but switched them to the interface types at the suggestion of @displagueto makes it easier to pass things in for testing and keep it standard.
> 
> Where do we do that testing?
> 
> Also, is it not legitimate to test if an interface type is nil?

I didn't see that type of testing done anywhere, so I think it's probably best to use the pointers for now unless we have a real need for it later, in which case we can introduce additional logic to check for nil where needed.

detiber: cc'ing @displague for visibilty

deitch: If you both are ok with it, I have no issues. @displague want to comment?

deitch: > Here is a better example of what I mean: https://go.dev/play/p/OkcY_Nvsc3H
> Since we are setting the value with the returned value of a constructor, the type of the argument at runtime is not actually nil

In your example, `f` actually is `nil`, as it never was set, but `h` returns from the constructor, so it is not `== nil`, but still will give an error if you try to call something on it. So basically, `== nil` does not catch all of the cases where we might have that issue. Is that it?

Do we actually trip this up somewhere in CPEM? I assume we do, or you wouldn't trouble yourself with this PR. Where is it?

deitch: Got it. Yeah, then this is a much better way. Let's Gove @displague a chance to comment

detiber: @deitch Can we go ahead and push forward with this? If needed we can always come back around and re-address the issue if needed.

deitch: Yup, plenty of time for people to respond, and we are not yet cutting a release anyways.

deitch (approved):

PR #234, opened 2022-03-18
Refactor test and release workflows

detiber:
- Add support for multiple registries
- conditionalize registry logins and image pushes
- add dockerignore
- Cleanup image creation
- bump golangci-lint
- use golangci-lint for vet, use golangci github action for lint
- Add caching for ci/release workflows
- breakout lint, test, image as separate jobs
- cleanup makefile a bit more
- Add dependabot config
- Skip ci workflow on dependabot branches
- Update default image used by helm chart and deployment.yaml

Signed-off-by: Jason DeTiberus <detiber@users.noreply.github.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,11 @@
+.git
+.github
+.vscode
+.golangci.yml
+deploy/
+dist/
+docs/
+**/.md
+*.md
+.gitignore
+Makefile
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  # Maintain dependencies for Go modules.
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+
+  # Maintain dependencies for Docker images.
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,8 +101,12 @@ jobs:
       id: should-push
       run: |
         SHOULD_PUSH=true
+        # Skip pushing if initiated by dependabot
         [[ "$GITHUB_ACTOR" == "dependabot[bot]" ]] && SHOULD_PUSH=false
+        # Skip pushing if a PR
         [[ "$GITHUB_EVENT_NAME" == "pull_request" ]] && SHOULD_PUSH=false
+        # Skip pushing if a pull request and branch is not master
+        [[ "$GITHUB_EVENT_NAME" == "push" && "$GITHUB_REF_NAME" != "master" ]] && SHOULD_PUSH=false
         echo "::set-output name=should-push::${SHOULD_PUSH}"
 
     - name: Log in to ghcr.io
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,6 +102,7 @@ jobs:
       run: |
         SHOULD_PUSH=true
         [[ "$GITHUB_ACTOR" == "dependabot[bot]" ]] && SHOULD_PUSH=false
+        [[ "$GITHUB_EVENT_NAME" == "pull_request" ]] && SHOULD_PUSH=false
         echo "::set-output name=should-push::${SHOULD_PUSH}"
 
     - name: Log in to ghcr.io
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,11 +4,11 @@ on:
     types: [opened, synchronize, reopened]
   push:
     branches:
-    - master
+    - '*'
+    - '!dependabot/**'
 
 env:
-  IMAGE_NAME: "equinix/cloud-provider-equinix-metal"
-
+  IMAGE_NAME: ${{ github.repository }}
 jobs:
   report:
     name: Report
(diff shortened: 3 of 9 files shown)


detiber:
Examples of these workflows running can be found here: https://github.com/detiber/packet-ccm/actions
Example releases can be found here: https://github.com/detiber/packet-ccm/releases

Note: only a few of the example releases have meaningful notes because there were no PRs with the exception of the dependabot PRs...

.dockerignore
```diff
@@ -0,0 +1,11 @@
+.git
```
detiber, line 1: This is to help with image caching, otherwise any file changes will cause a cache miss.

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
```
detiber, line 1: Let dependabot do some of the legwork for us when there are updates to dependencies.

.github/workflows/ci.yaml
```diff
@@ -4,11 +4,11 @@ on:
     types: [opened, synchronize, reopened]
   push:
     branches:
-    - master
+    - '*'
+    - '!dependabot/**'
```
detiber, line 8: Might as well run these for all branches, excluding dependabot because dependabot created branches do not have access to repo secrets.

.github/workflows/ci.yaml
```diff
@@ -4,11 +4,11 @@ on:
...
 
 env:
-  IMAGE_NAME: "equinix/cloud-provider-equinix-metal"
-
+  IMAGE_NAME: ${{ github.repository }}
```
detiber, line 11: This allows for forks to also be able to build images and use the workflows.

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+    steps:
+    - name: checkout
+      uses: actions/checkout@v3
+    - uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v3.1.0
+      with:
+        version: v1.44.2
```
detiber, line 33: Separate lint job to help speed up runs, the golangci-lint-action also provides additional caching and optimization beyond running it through make.

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+    - name: Go Build Cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ steps.go-cache-paths.outputs.go-build }}
+        key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-build-
+
+    - name: test
+      run: make race
```
detiber, line 65: Separate job for running the unit tests, with added caching.

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+
+  image:
```
detiber, line 67: Separate job for building and pushing the image

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+    - name: Build image urls
+      id: image-urls
+      env:
+        DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      run: |
+        IMAGES=ghcr.io/${IMAGE_NAME}
+        [[ -n "$DOCKER_USERNAME" ]] && IMAGES=${IMAGES},${IMAGE_NAME}
+        [[ -n "$QUAY_USERNAME" ]] && IMAGES=${IMAGES},quay.io/${IMAGE_NAME}
+        echo "::set-output name=images::${IMAGES}"
```
detiber, line 87: By default use ghcr.io, since we know we have credentials locally for that. Only add docker and quay.io registries when there are secrets available for them.

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+
+    - name: Docker manager metadata
+      id: meta
+      uses: docker/metadata-action@v3
```
detiber, line 91: docker/metadata-action automates quite a few different ways to generate tags based on what context the job is running in.

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+          type=ref,event=branch
+          type=ref,event=tag
+          type=ref,event=pr
+
+    - name: Determine if workflow should push image
+      id: should-push
+      run: |
+        SHOULD_PUSH=true
+        [[ "$GITHUB_ACTOR" == "dependabot[bot]" ]] && SHOULD_PUSH=false
+        echo "::set-output name=should-push::${SHOULD_PUSH}"
```
detiber, line 105: Gives a way to skip pushing if running on a dependabot-based branch, since it will have a read-only github token

.github/workflows/ci.yaml
```diff
@@ -18,44 +18,131 @@ jobs:
...
+          BINARY=cloud-provider-equinix-metal
+        tags: ${{ steps.meta.outputs.tags }}
+        labels: ${{ steps.meta.outputs.labels }}
         platforms: linux/amd64,linux/arm64
-        push: true
-        tags: |
-          ${{env.IMAGE_NAME}}:${{github.sha}}
-          ${{env.IMAGE_NAME}}:master
+        cache-from: type=gha, scope=${{ github.workflow }}
+        cache-to: type=gha, mode=max, scope=${{ github.workflow }}
```
detiber, line 148: Caching for building images. This won't save time for initial builds with new code, but saves significant time on repeated builds where non-code changes are made.

.github/workflows/release.yaml
```diff
@@ -5,69 +5,128 @@ on:
...
+        labels: ${{ steps.meta.outputs.labels }}
+        platforms: linux/amd64,linux/arm64
+        cache-from: type=gha, scope=${{ github.workflow }}
+        cache-to: type=gha, mode=max, scope=${{ github.workflow }}
 
   release: # create a github actions release with necessary artifacts
     name: Release
     runs-on: ubuntu-latest
+    needs:
+    - images
```
detiber, line 88: Add a dependency on the images job, so that the release is not present until the referenced image is created and pushed.

.github/workflows/release.yaml
```diff
@@ -5,69 +5,128 @@ on:
...
-      run: cat deploy/template/deployment.yaml | sed 's/RELEASE_TAG/'${{ steps.tagname.outputs.tag }}'/g' > /tmp/deployment.yaml
-    - name: Create Release
-      id: create_release
-      uses: actions/create-release@v1
+      env:
+        RELEASE_IMG: ghcr.io/${{ env.IMAGE_NAME }}:${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+      run: cat deploy/template/deployment.yaml | sed 's#RELEASE_IMG#${{ env.RELEASE_IMG }}#g' > /tmp/deployment.yaml
+    - name: Generate Release Notes
+      run: |
+        release_notes=$(gh api repos/{owner}/{repo}/releases/generate-notes -F tag_name=${{ github.ref }} --jq .body)
```
detiber, line 117: Get automated release notes form the github api

.golangci.yml
```diff
@@ -0,0 +1,10 @@
+linters:
+  disable-all: true
+  enable:
+  - gofmt
+  - govet
+  - revive
```
detiber, line 6: These are the pre-existing linters that we were running in the make file (revive is the successor to golint).

Dockerfile
```diff
@@ -1,30 +1,50 @@
-FROM alpine:3.15 as certs
-
-RUN apk --update add ca-certificates
-
-# builder
-FROM golang:1.17-alpine3.15 as build
-
-WORKDIR /go/src/app
-RUN apk --update add make git
+# syntax=docker/dockerfile:1.1-experimental
```
detiber, line 1: This is the same Dockerfile that we use in the cluster-api project, the use of distroless/static:nonroot will allow us to avoid copying the cert files from the alpine image which greatly improves caching.

Makefile
```diff
@@ -94,39 +90,24 @@ tag: ## Report the git tag that would be used for the images
...
-fmt-check: ## Check all source code formatting
-	@if [ -n "$(shell $(BUILD_CMD) gofmt -l ${GO_FILES})" ]; then \
-	  $(BUILD_CMD) gofmt -s -e -d ${GO_FILES}; \
-	  exit 1; \
-	fi
-
-fmt:   ## Format all source code files
-	$(BUILD_CMD) gofmt -w -s ${GO_FILES}
+fmt: golangci-lint  ## Format all source code files
+	@$(BUILD_CMD) $(LINTER) run --fix ./
```
detiber, line 94: standardize on golangci-lint, this also attempts to auto fix any issues similar to the old fmt target.

Makefile
```diff
@@ -94,39 +90,24 @@ tag: ## Report the git tag that would be used for the images
...
-	mkdir -p hacks && cd hacks && (go mod init hacks || true) && go get github.com/golangci/golangci-lint/cmd/golangci-lint@v1.27.0
-
-golint:
-ifeq (, $(shell which golint))
-	go get -u golang.org/x/lint/golint
-endif
-
-lint: pkgs golangci-lint ## Lint the files
-	@$(BUILD_CMD) $(LINTER) run --disable-all --enable=golint ./ ./metal
+	go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.44.2
```
detiber, line 98: Update to use the latest version of golangci-lint, since 1.27.0 is quite a bit old.

Makefile
```diff
@@ -169,7 +150,7 @@ sub-image-%:
 	@$(MAKE) ARCH=$* image
 
 image: ## make the image for a single ARCH
-	docker buildx build --load -t $(TAGGED_ARCH_IMAGE) -f Dockerfile --platform $(OS)/$(ARCH) .
+	docker buildx build --load --build-arg LDFLAGS="$(LDFLAGS_ARGS)" -t $(TAGGED_ARCH_IMAGE) -f Dockerfile --platform $(OS)/$(ARCH) .
```
detiber, line 153: plumb through ldflags to the image build, so that we get the expected version info back correctly.

detiber:
Looks like I need to bump up the golangci-lint timeout a bit more.

It also looks like I need to not attempt to push the built image when it's a PR (which makes sense given PRs don't have access to the repo secrets).

.dockerignore
deitch: Funny, I had through docker automatically ignores that. Go figure.

.github/workflows/ci.yaml
deitch: Nice. When it matches, this is much easier.

.github/workflows/ci.yaml
deitch: Should this not depend on the results of tests and lint? We only want to push out if build, test, lint pass. And only if this is merged to master or tagged, not PRs on other branches.

Dockerfile
deitch: Does it really make much of a difference?

deitch:
This is quite the cleanup! Thanks for tackling it. I only had a few questions, mostly marked inline.

.github/workflows/ci.yaml
detiber: Currently this replaces the previous build job, and pushing is disabled for PRs and dependabot branches which will lack access to the secrets in order to push, so this will just verify that the image (and source) can be built successfully in those cases.

We can exclude branches other than `master`, but I would argue that we could benefit from pushing images for other branches for easier testing purposes, and anyone pushing to a branch on this repo could be considered trusted from a secrets perspective.

Tags will be excluded here because we are specifying branches to match against, as can be seen from the workflows being run against my fork: https://github.com/detiber/packet-ccm/actions

Dockerfile
detiber: From a caching perspective, yes, since the `apk --update` commands invalidate the cache at the very beginning of the build. It took me a bit to figure out why I wasn't seeing any benefits of caching after adding the .dockerignore and verifying that the build context was unchanged between runs 😬

.golangci.yml
detiber: Over time we can go ahead and add additional linters as well, which would help highlight issues like in #232 where you caught an issue with my `UpdateFunc` in review, where I wasn't using `old` and instead was using `obj` twice 😬

deitch: 😄

Dockerfile
deitch: Well, if it works it works.

.github/workflows/ci.yaml
deitch: It isn't just a question of what has the right to push, although that is a big issue. I also don't think we want images out there and publicly available for anything other than what has been approved and merged. People can end up using them accidentally, and there may even be ownership/responsibility/liability issues.

Let's do the following. We currently only ever push out for merged-to-master and semver-tagged. Let's keep it this way, so that this PR focuses on all of the goodness we gain without changing behaviour. Then, if you want, open a separate PR that just pushes those out, and we can focus the discussion there.

Does that make sense?

detiber:
> This is looking pretty impressive. It needs a rebase, and I think we should leave the pushed images solely for `master` and semver tags for now. We always can open a PR to focus just on that discussion in the future.

Updated for at least the push/pr based job, it will now only build the images when it's a push and when branch is 'master'.

For the tag-based release job, right now it only enforces that the tag start with a `v`, not that the tag specifically relates to a semantic version, but this was the previous behavior, so it's not a regression in behavior.

deitch:
I changed the settings to require your new job names instead of the old ones. Looks good now.

PR #235, opened 2022-03-21
document requirements for e2e tests

deitch:
Document requirements for end-to-end tests. Currently, this is how I do the tests manually. With a clear list in a shared space, we should be able to plan towards automating them as well.


displague (approved):
Capturing the current state is a no-brainer. We can improve upon this over time.

deitch:
This is the way.

PR #248, opened 2022-04-03
Fix reading of annotation names from config file

jvassev:
The only way to override annotations is using env. The fix also
simplifies config param overriding


deitch:
This looks pretty good. I like the `override()` cleanup, makes things much easier to read and understand. Let's let CI do its thing.

deitch (approved):

PR #249, opened 2022-04-04
fix metallb matchlabels to be valid

### LINKED ISSUE #247: metallb `nomatch` labels violate valid label spec
In the migration to [LoadBalancer](https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer)  spec that we did in #224 , we had to find some way for metallb to track what services are served by each upstream peer-local pairing. This allowed us to then update or remove a pairing (as expected by the cloud-provider [LoadBalancer](https://pkg.go.dev/k8s.io/cloud-provider#LoadBalancer) API.

However, metallb provides no way of tracking it, nor does it have any additional comments or unused fields in its control configmap. We eventually settled on using the `match-labels` part of the metallb configmap.

For example:

```yaml
- my-asn: 65000
  peer-asn: 65530
  peer-address: 169.254.255.1
  peer-port: 0
  source-address: 10.68.190.5
  hold-time: ""
  router-id: ""
  node-selectors:
  - match-labels:
      kubernetes.io/hostname: dc-worker-1
    match-expressions: []
  - match-labels:
      nomatch.metal.equinix.com/service-name: default/nginx-deployment
    match-expressions: []
```

Since `node-selectors` is logical OR, the label `nomatch.metal.equinix.com/service-name` never will match anything, and so will be ignored by metallb, while allowing us to use it to track services. It also prevents us from having to have a pairing entry for each service, which would make the configmap grow to an untenable size very quickly (and probably break metallb).

However, this is invalid:

```yaml
  - match-labels:
      nomatch.metal.equinix.com/service-name: default/nginx-deployment
```

If we read the [description of valid characters](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set), it says:

> Valid label value:
> 
> * must be 63 characters or less (can be empty),
> * unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
> * could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

That means that `nomatch.metal.equinix.com/service-name` is perfectly fine as a key, but `default/nginx-deployment` (representing the service `nginx-deployment` in the namespace `default`) is not.

This now is breaking metallb.

Thanks to @enkelprifti98 for finding this.

We need a solution.

Some possibilities:

* replace the `/` we conventionally use to separate namespace from service, with some other valid character. Unfortunately, almost all valid characters also might be part of the namespace or service. Would `default_nginx_deployment` represent the deployment `nginx_deployment` in `default`, or `deployment` in `nginx_default`?
* use the `.` to separate them, as in `default.nginx-deployment`. This only works if `.` is not valid in either namespace or service name. I'm not sure of that. The page for naming is [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/), but it isn't clear what constraints namespaces and services have.
* hash or base64 the whole thing. Since we always know what we are comparing against, any consistent forward hash/base64 should work. E.g. instead of `nomatch.metal.equinix.com/service-name: default/nginx-deployment` we could do `nomatch.metal.equinix.com/service-name-base64: ZGVmYXVsdC9uZ2lueC1kZXBsb3ltZW50`. It is ugly, but it is usable and predictable, and has precedent in kubernetes' secrets
* other ideas

Thoughts @displague @cprivite @detiber @ocobleseqx ?

deitch:
fixes #247 
cc @displague @enkelprifti98

### DIFF
--- a/metal/loadbalancers/metallb/configmap.go
+++ b/metal/loadbalancers/metallb/configmap.go
@@ -57,7 +57,7 @@ func (cfg *ConfigFile) AddPeer(add *Peer) bool {
 // If a matching peer already exists with the service, do not change anything.
 // If a matching peer already exists but does not have the service, add it.
 // Returns if anything changed.
-func (cfg *ConfigFile) AddPeerByService(add *Peer, svc string) bool {
+func (cfg *ConfigFile) AddPeerByService(add *Peer, svcNamespace, svcName string) bool {
 	var found bool
 	// ignore empty peer; nothing to add
 	if add == nil {
@@ -70,16 +70,19 @@ func (cfg *ConfigFile) AddPeerByService(add *Peer, svc string) bool {
 	// - ASN matches
 	// - Addr matches
 	// - NodeSelectors all match (but order is ignored)
+	var peers []Peer
 	for _, peer := range cfg.Peers {
 		if peer.EqualIgnoreService(add) {
 			found = true
-			peer.AddService(svc)
+			peer.AddService(svcNamespace, svcName)
 		}
+		peers = append(peers, peer)
 	}
+	cfg.Peers = peers
 	if found {
 		return true
 	}
-	add.AddService(svc)
+	add.AddService(svcNamespace, svcName)
 	cfg.Peers = append(cfg.Peers, *add)
 	return true
 }
@@ -104,14 +107,14 @@ func (cfg *ConfigFile) RemovePeer(remove *Peer) {
 // For any peers that have this services in the special MatchLabel, remove
 // the service from the label. If there are no services left on a peer, remove the
 // peer entirely.
-func (cfg *ConfigFile) RemovePeersByService(svc string) bool {
+func (cfg *ConfigFile) RemovePeersByService(svcNamespace, svcName string) bool {
 	var changed bool
 	// go through the peers and see if we have a match
 	peers := make([]Peer, 0)
 	// remove that one, keep all others
 	for _, peer := range cfg.Peers {
 		// get the services for which this peer works
-		peerChanged, size := peer.RemoveService(svc)
+		peerChanged, size := peer.RemoveService(svcNamespace, svcName)
 
 		// if not changed, or it has at least one service left, we can keep this node
 		if !peerChanged || size >= 1 {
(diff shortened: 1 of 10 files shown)


displague:
Let's add an example to https://github.com/equinix/cloud-provider-equinix-metal#metallb

> CCM itself does not deploy the load-balancer or any part of it, including the ConfigMap. It only modifies an existing ConfigMap. This can be deployed by the administrator separately, using the manifest provided in the releases page, or in any other manner.

Maybe something like this:

The CCM modifies the MetalLB ConfigMap with special node-selectors. These node-selectors are ignored by MetalLB but are critical to the CCM management of MetalLB configuration. For example (perhaps we can make this look more practical):
```
- my-asn: 65000
  peer-asn: 65530
  peer-address: 169.254.255.1
  peer-port: 0
  source-address: 10.68.190.5
  hold-time: ""
  router-id: ""
  node-selectors:
  - match-labels:
      kubernetes.io/hostname: dc-worker-1
  - match-labels:
      nomatch.metal.equinix.com/service-namespace: default
      nomatch.metal.equinix.com/service-name: nginx-deployment
 - match-labels:
      nomatch.metal.equinix.com/service-namespace: ai
      nomatch.metal.equinix.com/service-name: trainer
 - match-labels:
      nomatch.metal.equinix.com/service-namespace: database
      nomatch.metal.equinix.com/service-name: mysql1
```

metal/loadbalancers/metallb/metallb.go
```diff
@@ -15,10 +15,11 @@ import (
 )
 
 const (
-	hostnameKey      = "kubernetes.io/hostname"
-	serviceNameKey   = "nomatch.metal.equinix.com/service-name"
-	defaultNamespace = "metallb-system"
-	defaultName      = "config"
+	hostnameKey         = "kubernetes.io/hostname"
+	serviceNameKey      = "nomatch.metal.equinix.com/service-name"
+	serviceNameSpaceKey = "nomatch.metal.equinix.com/service-namespace"
```
displague, line 20: Space is vast.

```suggestion
	serviceNamespaceKey = "nomatch.metal.equinix.com/service-namespace"
```

metal/loadbalancers/metallb/configmap.go
```diff
@@ -484,51 +487,67 @@ func (p *Peer) EqualIgnoreService(o *Peer) bool {
...
-func (p *Peer) Services() []string {
+func (p *Peer) Services() []Resource {
+	var services []Resource
 	for _, ns := range p.NodeSelectors {
+		var name, namespace string
 		for k, v := range ns.MatchLabels {
-			if k == serviceNameKey {
-				return strings.Split(v, ",")
+			switch k {
+			case serviceNameKey:
```
displague, line 496: What are your thoughts on _healing_ the old format?  We would detect when `serviceNameKey` contains the old format and append any/all services detected.

I presume we would need to build a similar capability into the other for loops walking nodeselectors (which makes me wonder if these could be more consolidated).

metal/loadbalancers/metallb/metallb.go
deitch: It is, indeed, the final frontier. We can do this.

metal/loadbalancers/metallb/configmap.go
deitch: Is there any point? Given that the old format is broken, and shouldn't work, why bother?

deitch:
Updated the README. Ready for another look @displague

displague (approved):

metal/loadbalancers/metallb/configmap.go
displague: Fair.

PR #250, opened 2022-04-05
Update automated workflow to allow for overriding orgs for dockerhub and quay

detiber:
Update automated workflow to allow for overriding orgs for dockerhub and quay

### DIFF
--- a/.github/actions/metadata/action.yaml
+++ b/.github/actions/metadata/action.yaml
@@ -1,0 +1,46 @@
+name: 'Metadata'
+description: 'Generate Image Metadata'
+inputs:
+  docker_username:
+    description: 'docker_username'
+    required: true
+  quay_username:
+    description: 'quay_username'
+    required: true
+  metadata_flavor:
+    description: 'metadata_flavor'
+    required: true
+  metadata_tags:
+    description: 'metadata_tags'
+    required: true
+outputs:
+  tags:
+    description: "generated image tags"
+    value: ${{ steps.meta.outputs.tags }}
+  labels:
+    description: "generated image labels"
+    value: ${{ steps.meta.outputs.labels }}
+  version:
+    description: "generated image version"
+    value: ${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+runs:
+  using: "composite"
+  steps:
+  - name: Build image urls
+    shell: bash
+    id: image-urls
+    env:
+      DOCKER_USERNAME: ${{ inputs.docker_username }}
+      QUAY_USERNAME: ${{ inputs.quay_username }}
+    run: |
+      IMAGES=${GHCR_REGISTRY}/${IMAGE_NAME}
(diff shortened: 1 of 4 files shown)


deitch:
@detiber a few questions:

* why do we have our own metadata action here? What does it do?
* why move the org names into secrets?

detiber:
> @detiber a few questions:
> 
> * why do we have our own metadata action here? What does it do?

It allows to standardize a bit how we are creating the metadata and avoiding some copy-pasta type issues, since it needs to be run by multiple jobs.

> * why move the org names into secrets?

I'd rather not use a secret here TBH, but it's the only way that I could figure out how to override the value on a per-org basis, so that users can run/publish/test using their own quay/dockerhub/ghcr orgs for testing in their own forks.

Otherwise users would have to hack things up and override hardcoded values in the repo in order to override upstream quay/dockerhub orgs.

deitch:
> I'd rather not use a secret here TBH, but it's the only way that I could figure out how to override the value on a per-org basis, so that users can run/publish/test using their own quay/dockerhub/ghcr orgs for testing in their own forks

So if I want to test the CI by launching an action manually, but pushing to a different org (not affect the official distributed supported EQXM images), then I can change the secret? But how would I do that without being admin, and doesn't it then permanently change it, so now I need an admin to put it back?

Admittedly, testing for actions is kind of miserable.

detiber:
> > I'd rather not use a secret here TBH, but it's the only way that I could figure out how to override the value on a per-org basis, so that users can run/publish/test using their own quay/dockerhub/ghcr orgs for testing in their own forks
> 
> So if I want to test the CI by launching an action manually, but pushing to a different org (not affect the official distributed supported EQXM images), then I can change the secret? But how would I do that without being admin, and doesn't it then permanently change it, so now I need an admin to put it back?
> 
> Admittedly, testing for actions is kind of miserable.

So, basically, the secrets only exist for this repository. If a user wanted to test on their fork, they would define their own secrets there. Since their fork is under their personal account, they have admin permissions to add secrets there.

detiber:
As a more concrete example, when testing this in my fork, I configured `DOCKER_ORG` to `detiber`, and `QUAY_ORG` to `detiber`, and that allowed me to test everything against container repos that I owned.

deitch:
Oh, I get it. So you can test on your own fork, then open the PR when the action looks good.

There _seriously_ has to be a better way to test actions. My instinct says, "oh, just push to a local registry", but that doesn't work when CI is running on GH Actions in their runners.

I guess it looks good, then, but do you mind adding something to the docs about testing, i.e. how to do what you just described? "in your own fork, add secrets, etc."

Then we need @displague to add the necessary secrets before merging.

detiber:
@deitch this should be ready to go now :)

deitch (approved):
LGTM. Still needs to wait for secrets from @displague ?

detiber:
> LGTM. Still needs to wait for secrets from @displague ?

Nope, I updated the secrets a bit ago after @displague sent them my way.

deitch:
Cool. Then let's do it!

Issue #257, opened 2022-04-21
build is long and slow


deitch:
CI takes about 30 mins to run, primarily one step. I don't believe this uses cgo, so what could it be building? We should see what it is doing, and if there are ways to speed it up.

deitch:
I did some digging into this. What I am seeing is that the vast majority of build steps take at most a few seconds, often less than a second.

A few steps take longer:

* `go mod download` for native platform, ~24 seconds.
* `go mod download` for alternate platform, i.e. under qemu, ~3 mins.
* `go build` for native platform, ~4 mins.
* `go build` for alternate platform, i.e. under qemu, 17-20 mins

Total time for the above: 17-20 mins +3 mins + 4 mins + 24 seconds = 24.5-27.5 mins

Clearly, the biggest issue is qemu emulated build. This isn't too surprising, given qemu speed. However, given that we are compiling a single go binary with `CGO_ENABLED=0`, we should be able to get a major performance boost by cross-compiling using go on the native platform, and using qemu only to assemble the final image. That should get our build to be closer to 10 mins. Not exactly clear how the actions support that, but we can look into it.

Once that is in place, it is worth questioning why the native `go build` takes 4 minutes. I assume it is because it needs to compile all of the library dependencies for the first time, whereas when running locally, I have all of the pkg dependences already compiled, and just the actual CPEM code needs to be compiled again.

deitch:
#260 now is open to handle the cross-compiling natively. After that, we can shave a few minutes off by caching the `go mod download`.

deitch:
Fixed by #260

PR #260, opened 2022-04-24
use native go cross-compiling

deitch:
We call the builds by passing `--platforms` to the `docker buildx build`, whether directly or via the GitHub action `docker/build-push-action`.

Either way, this causes it to run every command under qemu, so that non-native builds take about 4-5x as long as the native ones. See a typical build and issue #257 .

This proposes to fix the biggest problem, i.e. those emulated builds. Instead of emulating the entire process, this will run the actual build (as well as `go mod download` and every other step except for image assembly) on the native platform, but cross-compiling.

Local tests show that the result is that building for both native and alternate platforms are within a few percentage points of each other. We will let CI run here and see what it reports.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,10 @@
 
 # Build the manager binary
 ARG GOVER=1.17.8
-FROM golang:${GOVER} as builder
+FROM --platform=$BUILDPLATFORM golang:${GOVER} as builder
+
+ARG TARGETPLATFORM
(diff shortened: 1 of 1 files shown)


deitch:
Build took 5.5 minutes, vs 20-25 mins before. I think we can consider this a win.

displague (approved):
Simple and large optimizations are the best!

PR #261, opened 2022-04-28
fix ghcr.io org name for releases

deitch:
We were using the `QUAY_ORG` instead of the ghcr org (which, by definition, is the `repository_owner`), so no one could pull it down.


displague:
Looks like we also need to update `ci.yaml`.

displague (approved):

Issue #296, opened 2022-06-20
Support VLANs with customer-managed IPs


deitch:
CPEM currently supports only EQXM-provided IPs. It relies on retrieving the public and private IPs for each device from the EQXM APIs. This means that customers using VLANs, with IPs managed by them, are not supported properly, as the CCM does not know about them.

We need to add support for VLANs, or more generally customer-managed IPs.

deitch:
This was raised in #193 specifically around one use case, but led to a generic conversation around support for these IPs beginning [here](https://github.com/equinix/cloud-provider-equinix-metal/issues/193#issuecomment-1111177645) with input from @displague .

deitch:
The way it works is that `kubernetes/cloud-provider` calls `InstanceMetadata()` in the CCM to get information about the `v1.Node`, including the correct "internal" (to be used for intra-cluster communication) and "external" to be used for outside processes to connect to the node) IPs. It provides a list of IPs for each, the first of which for each type is the "official" one.

Currently, CCM [here](https://github.com/equinix/cloud-provider-equinix-metal/blob/8c363dab030d0bf8424139aae946fd316a14a145/metal/devices.go#L99-L127) simply takes the entire list from the EQXM API, sorts them into public-as-external and private-as-internal, and returns them.

The proposed solution is:

* if no existing node annotation indicates that an IP was provided, continue to use the current logic
* if a node annotation indicates that an IP was provided, use that as the first internal IP, followed by all others per the current logic
* if the CCM was started with the flag `--provided-node-ips-only` (open to a better name), or its parallel config file flag or env var, then it will _only_ use the provided IP, both for internal and external. If none is provided, return an error.

This enables 3 scenarios:

* I don't have my own IP management (L3): do not set the CCM flag, do not override the node-ip with the kubelet flag
* I have my own IPs but also want the existing EQXM ones (L2 hybrid): do not set the CCM flag, but do override the node-ip with the kubelet flag
* I have my own IPs and want only those and not the EQXM ones (L2): set the CCM flag, override the node-ip with the kubelet flag

PR #297, opened 2022-06-27
when --node-ip provided, use that as first InternalIP

deitch:
This provides full support to using VLAN IPs and floating IPs.


deitch:
Fixes #296

metal/devices.go
```diff
@@ -96,11 +103,25 @@ func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
...
+	if providedNodeIP != "" {
+		privateIP = providedNodeIP
+		addr = v1.NodeAddress{Type: v1.NodeInternalIP, Address: providedNodeIP}
+		if _, ok := unique[addr]; !ok {
+			unique[addr] = true
+			addresses = append(addresses, addr)
+		}
+	}
 	for _, address := range device.Network {
 		if address.AddressFamily == int(metadata.IPv4) {
```
displague, line 126:
```suggestion
		if providedNodeIP == address.Address {
			continue
		}
		if address.AddressFamily == int(metadata.IPv4) {
```

The `unique` map would have the public `providedNodeIP` address indexed as `NodeInternalIP` and if we don't do something like the above we'll add a second index for the same address as `NodeExternalIP`.

(GKE permits public internalips too https://cloud.google.com/kubernetes-engine/docs/how-to/alias-ips#enable_pupis)

metal/devices.go
```diff
@@ -111,7 +132,12 @@ func nodeAddresses(device *packngo.Device) ([]v1.NodeAddress, error) {
 				privateIP = address.Address
 				addrType = v1.NodeInternalIP
 			}
-			addresses = append(addresses, v1.NodeAddress{Type: addrType, Address: address.Address})
+			addr = v1.NodeAddress{Type: addrType, Address: address.Address}
+
+			if _, ok := unique[addr]; !ok {
```
displague, line 137: We could take advantage of this CloudProvider helper and avoid some of the logic here

https://github.com/kubernetes/cloud-provider/blob/9a26d4868f3a005a0cb22e19c189499642a29b04/node/helpers/address.go#L27-L42

metal/devices.go
```diff
@@ -60,13 +61,19 @@ func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 	if err != nil {
 		return nil, err
 	}
-	nodeAddresses, err := nodeAddresses(device)
+	// was a node IP provided
+	var providedNodeIP string
+	if node.Annotations != nil {
+		providedNodeIP = node.Annotations[cpapi.AnnotationAlphaProvidedIPAddr]
+	}
+	nodeAddresses, err := nodeAddresses(device, providedNodeIP)
```
displague, line 69: I think we would get this for free by starting with `nodeAddresses := append([]v1.NodeAddress, node.Status.Addresses...)` (however it is we get the existing addresses).

If we don't use the existing node addresses, we may omit addresses manually defined or defined through other kubelet options. For example `--hostname-override`:
* https://kubernetes.io/docs/reference/labels-annotations-taints/#kubernetesiohostname
* https://kubernetes.io/docs/concepts/architecture/nodes/#addresses

metal/devices.go
```diff
@@ -60,13 +61,19 @@ func (i *instances) InstanceMetadata(ctx context.Context, node *v1.Node) (*cloud
 	if err != nil {
 		return nil, err
 	}
-	nodeAddresses, err := nodeAddresses(device)
+	// was a node IP provided
+	var providedNodeIP string
+	if node.Annotations != nil {
+		providedNodeIP = node.Annotations[cpapi.AnnotationAlphaProvidedIPAddr]
```
displague, line 67: I noticed something interesting about `--node-ip` behavior.

https://github.com/kubernetes/cloud-provider/blob/9a26d4868f3a005a0cb22e19c189499642a29b04/node/helpers/address.go#L92-L110

> // * If nodeIP matches an address of a particular type (internal or external),
> //   that will be the *only* address of that type returned.

metal/devices.go
deitch: Maybe it makes more sense to have the `unique` map index on the address string, rather than the `v1.Address` struct? That way it guarantees uniqueness?

```go
// partial code
var 		unique              = map[v1.NodeAddress]bool{}
	addr := v1.NodeAddress{Type: v1.NodeHostName, Address: device.Hostname}
	unique[addr.Address] = true
// etc.
```

metal/devices.go
deitch: Instead of the whole `unique` thing? I guess so. I don't really care that much. It replaces 4 lines of code with 1 line of code 3 times, which really isn't all that much. It also is inefficient, looping each time (which doesn't matter all that much in the scale of a few entries). But it also eliminates our ability to do the `unique` change I suggested above to fit your requirements.

metal/devices.go
deitch: Done.

metal/devices.go
deitch: There are no other structures to provide addresses to the kubelet other than `--node-ip` and `--hostname-override`. This PR addresses `--node-ip`, but I think you are  correct to raise that it does _not_ resolve the issue of `--hostname-override`. It actually explicitly sets the hostname based on the response from the EQXM API.

I am going to leave this one for now. It isn't quite so simple. There are 3 distinct hostnames to worry about:

*  whatever the kubelet got from the node automatically (by running `hostname`); this might to might not be what is in the EQXM API, e.g. the user changed the hostname via command manually, userdata script, etc.
* `--hostname-override`
* EQXM API

Reconciling all three of those is its own logical mess. So I am leaving it out of this PR. This PR doesn't change the logic, however flawed: hostname is set via EQXM API.

We can set a new one for hostname after this one is done.

metal/devices.go
displague: I don't see this change committed.

I would expect to find:
```go
var unique = map[string]bool{}
addr := v1.NodeAddress{Type: v1.NodeHostName, Address: device.Hostname}
unique[addr.Address] = true
```

But instead I see: https://github.com/equinix/cloud-provider-equinix-metal/pull/297/files#diff-3ddaf4333cb941b353834040ab733477c56a207db5bcf2e854b62fe663db5494R109-R113

(maybe I'm github'ing incorrectly)

deitch: Aha! I had a failed push due to bad network, which I hadn't realized. Pushed now and updates.

deitch: > githubing

Is that a verb now? 😁

displague (approved):

Issue #304, opened 2022-07-29
Improve EIP/VIP failover when the node running CPEM fails


cprivitere:
Right now in a cluster with an HA control plane using CPEM for EIP management you can have the following situation.

### Situation
3-controlplane-node Kubernetes cluster, created with CAPP. Has CPEM, with the control plane EIP pointing at one of the nodes (not kube-vip or metallb, but rather CPEM’s direct control), CPEM running on the same node that EIP is pointing to.
CP nodes
  A    <--- EIP ; CPEM here
  B
  C (edited) 

Node A goes away.

CPEM should move the EIP to node B or C. But CPEM is gone too, so it cannot do so until k8s controllers move it to node B.
Problem: It never happens. Controller and scheduler never seem to learn that it has something to do, CPEM stays dead, EIP never moves, we are stuck.
What does work: Each CP node’s controller-manager and scheduler do point at their local IP kube-apiserver, which in turn points at the local etcd.

###Root cause
kubelet. /etc/kubernetes/kubelet.conf points to the EIP:
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data:  // etc. etc.
    server: https://1.2.3.4:6443/      // <===== this is the EIP, not the local
  name: default-cluster
```

So all of the components are working, but the kubelets cannot connect to the apiserver, so controllers do not actually know what the state of anything is.

cprivitere:
I did investigation into what options cluster-api lets you pass down to kubeadm configuration for changing what IPs the kubelet is configured to use. Not only is there nothing that lets you override the address kubelet talks to with cluster-api OR kubeadm, I eventually found this:

**Note**: kubeadm applies the same KubeletConfiguration to all nodes in the cluster. To apply node specific settings you can use kubelet flags as overrides by passing them in the nodeRegistration.kubeletExtraArgs (i’ve checked, there are no flags that help us) field supported by both InitConfiguration and JoinConfiguration. Some kubelet flags are deprecated, so check their status in the [kubelet reference documentation](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet) before using them.

So, the long and short is that my research is indicating that overriding the IP that kubelet is set to talk to just for the controplanes isn’t really a normal pattern (at least not for kubeadm).

I’m concerned we need to look more at changing how CPEM runs or just go with a hack that plumbs the main EIP onto each control planes localhost interface.

cprivitere:
[...]

PR #310, opened 2022-08-19
switch to DaemonSet to run on all control plane nodes

deitch:
Partially addresses #304, specifically the second issue raised.

Rather than running it as a `Deployment` with `replicas=1`, we run this as a `DaemonSet` with `nodeAffinity` restricting it to control plane nodes. This means that if you have 3 such nodes, you will get 3 copies.

The built-in leader election of k8s.io/cloud-provider ensures that only one processes events at a time.

If you have multiple (as this provides), and one dies (of greater concern, the node it is running on dies, taking CPEM with it), then one of the remaining ones will take over quickly, and work will continue as normal. This includes the important part of letting the apiserver know that the node is gone, which is a CPEM responsibility.

**Important note::** This does _not_ solve the issue of the node that dies being host both to the CPEM (whether lone before this PR or current leader after this PR) _and_ the EIP (when using an EIP managed by CPEM for apiserver access. In that case, this will not help. As the node goes down, so does CPEM, so nothing can switch the EIP to a functioning node. That is CPEM's responsibility, but it is down, too. Leader election would help, but leader election depends on access to the k8s apiserver, which depends on the EIP, which points to the node that just went down.

That will need to be addressed via some other solution; see the tracking issue linked at the beginning.

---
Separately, while also dealing with the deployment templates, it also fixes a deprecated annotation. This used to exist:

```yaml
      annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
```

But has been deprecated since 1.16. The replacement to use:

```yaml
priorityClassName: system-cluster-critical
```

as part of the podspec has been adopted.

### DIFF
--- a/deploy/chart/templates/deployment.yaml
+++ b/deploy/chart/templates/deployment.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: DaemonSet
 metadata:
   name: {{ include "cloud-provider-equinix-metal.fullname" . }}
   labels:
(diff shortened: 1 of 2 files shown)


deitch:
Still doing some testing, so hang on with it.

deitch:
This is good to go. Looking for a review

cprivitere (approved):

PR #318, opened 2022-08-29
bump to go 1.19, which is a prerequisite for many dependency upgrades

deitch:

PR #320, opened 2022-08-29
k8s.io/* to v0.25.0

deitch:
With go-1.19 and lint fixes in #318 , we are ready for the bump of k8s.io v0.25.0. 

This also will automatically take care of 4 dependenabot ones.


deitch:
Had to rerun, but CI clean now.

PR #367, opened 2023-02-22
Change-default-branch-name-to-main

cprivitere:
This PR would only be merged once the default branch name was switched to main.

Signed-off-by: Chris Privitere <23177737+cprivitere@users.noreply.github.com>


README.md
```diff
@@ -13,7 +13,7 @@
 
 `cloud-provider-equinix-metal` is the Kubernetes CCM implementation for Equinix Metal. Read more about the CCM in [the official Kubernetes documentation](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/).
 
-This repository is [Maintained](https://github.com/packethost/standards/blob/master/maintained-statement.md)!
+This repository is [Maintained](https://github.com/packethost/standards/blob/main/maintained-statement.md)!
```
displague, line 16: nit: this could be more rewritten to skip a redirect https://github.com/equinix-labs/equinix-labs/blob/main/maintained-statement.md

displague: Same with the other packethost/standards repo references

deploy/chart/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: cloud-provider-equinix-metal
 description: Kubernetes Cloud Provider for Equinix Metal 
 type: application
 version: "0.1.0"
-appVersion: "master"
+appVersion: "main"
```
displague, line 6: Are there `helm upgrade` implications to this change?

displague (approved):
I have one question about the `helm upgrade` path, but I don't believe Helm is a popular install mechanism for this project. We may need to call the appVersion change out in the release notes if it isn't seamlessly handled.

ctreatma (approved):

ctreatma:
Based on the [Chart.yaml docs for that field](https://helm.sh/docs/topics/charts/#the-appversion-field), I think the `appVersion` will not impact users:

> This field is informational, and has no impact on chart version calculations.

deploy/chart/Chart.yaml
cprivitere: I don't believe so as it's my understanding that this helm doesn't actually exist anywhere other than this chart file. It's just an experiment.

README.md
cprivitere: I've replaced the two instances I could find.

ctreatma (approved):

PR #368, opened 2023-02-22
Remove qemu and related items from Makefile

cprivitere:
Docker build (we're still using buildx, but close enough) no longer needs us to do any manual qemu setup for cross-architecture building. So I've removed the QEMU variables and the register: block from the Makefile.

Signed-off-by: Chris Privitere <23177737+cprivitere@users.noreply.github.com>


displague (dismissed):
buildx results look clean in CI

ctreatma (approved):

PR #386, opened 2023-03-06
Add explicit CRD configuration for MetalLB

### LINKED ISSUE #385: MetalLB pods can not be read to detect MetalLB version
Detecting MetalLB version by pod image is problematic. MetalLB pods can not be read to detect MetalLB version without proper RBAC. Pod image detection is tricky because namespaces, resource types, an resource names are all under the user's control.

Recommended approach is to have the user specify metallb version as a command line arg, environment variable, or configmap setting.

### LINKED ISSUE #387: CCM does not support the documented MetalLB config options
From a comment on #385:

Hi, I'm just testing latest commit of main branch (660ca1e102ee389ffd240d28c20954303a0ca01f)

I see different issue.

I configured `METAL_LOAD_BALANCER: "metallb:///kube-system"` (metallb is running in `kube-system`) according to docs.
```markdown
To enable it, set the configuration METAL_LOAD_BALANCER or config loadbalancer to:

metallb:///<configMapNamespace>

For example:

    metallb:///metallb-system - enable MetalLB management and update configuration in the namespace metallb-system (default)
    metallb:///foonamespace - enable MetalLB management and update configuration in the namespace metallb-system
    metallb:/// - enable MetalLB management and update configuration in the default namespace metallb-system
```

In the function `NewLB()` there is a splitting functionality. The value of `config` when function is started is `/kube-system`. First `/` is removed - value of `config` is `kube-system`. Then `cmparts` does basically nothing - so length of `cmparts` is 1. This way the `if len(cmparts) >= 2` is lower than 2 and namespace/configmapname is not populated.

Here is screenshot from debug session in VSCode:
Screenshot from 2023-03-07 09-56-21

Then since the `namespace` of MetalLB is wrong the code can't compare current/supported versions and doesn't use CRDs.

ctreatma:
Fixes #385 
Fixes #387 

The old MetalLB version detection code relied on permissions that the CCM doesn't have in order.  Instead, this adds support for an explicit `crdConfiguration` flag as a URL parameter in the `loadBalancer` configuration string.  Initially, the `crdConfiguration` flag defaults to `false` for backwards compatibility; we will switch the default in a future major release.

In addition, this fixes the MetalLB configuration code so that the documented namespace-only configuration (`metallb:///<namespace>`), which is essential for MetalLB 0.13.2+, is supported

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,10 +42,10 @@ Before committing, it's a good idea to run `goimports -w .`.
 
 ## Building and Testing
 
-The [Makefile](./Makefile) contains the targets to build, lint, vet and test:
+The [Makefile](./Makefile) contains the targets to build, lint, and test:
 
 ```sh
-make build lint vet test
+make build lint test
 ```
 
 These normally will be run using your locally installed golang tools. If you do not have them
@@ -53,7 +53,7 @@ installed, or do not want to use local ones for any other reason, you can run it
 image by setting the var `DOCKERBUILD=true`:
 
 ```sh
-make build lint vet test DOCKERBUILD=true
+make build lint test DOCKERBUILD=true
 ```
 
 If you want to see HTTP requests, set the `PACKNGO_DEBUG` env var to non-empty
--- a/README.md
+++ b/README.md
@@ -235,7 +235,6 @@ This section lists each configuration option, and whether it can be set by each 
 | Kubernetes API server port for Elastic IP |     | `METAL_API_SERVER_PORT` | `apiServerPort` | Same as `kube-apiserver` on control plane nodes, same as `0` |
 | Filter for cluster nodes on which to enable BGP |    | `METAL_BGP_NODE_SELECTOR` | `bgpNodeSelector` | All nodes |
 | Use host IP for Control Plane endpoint health checks | | `METAL_EIP_HEALTH_CHECK_USE_HOST_IP` | `eipHealthCheckUseHostIP` | false |
-| Use newer MetalLB (0.13.2+) configuration |   | `METAL_USE_CRD_FOR_METALLB` | `useCRDForMetalLB` | false |
 
 <u>Security Warning</u>
 Including your project's BGP password, even base64-encoded, may have security implications. Because Equinix Metal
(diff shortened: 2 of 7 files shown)


CONTRIBUTING.md
```diff
@@ -42,18 +42,18 @@ Before committing, it's a good idea to run `goimports -w .`.
 
 ## Building and Testing
 
-The [Makefile](./Makefile) contains the targets to build, lint, vet and test:
+The [Makefile](./Makefile) contains the targets to build, lint, and test:
```
ctreatma, line 45: Changes in this file are cleanup unrelated to the MetalLB issue; I can break this file out into a separate PR if needed.

README.md
```diff
@@ -214,7 +214,7 @@ This section lists each configuration option, and whether it can be set by each
 
 | Purpose | CLI Flag | Env Var | Secret Field | Default |
 | --- | --- | --- | --- | --- |
-| Path to config secret |    |    | `cloud-config` | error |
+| Path to config secret | `cloud-config` |    |  | error |
```
ctreatma, line 217: This change is a docs fix unrelated to the MetalLB issue.  I can break it out into a separate PR if needed.

metal/loadbalancers/metallb/metallb.go
```diff
@@ -65,7 +62,7 @@ var (
 )
 
 // func NewLB(k8sclient kubernetes.Interface, k8sApiextensionsClientset *k8sapiextensionsclient.Clientset, config string) *LB {
-func NewLB(k8sclient kubernetes.Interface, config string) *LB {
+func NewLB(k8sclient kubernetes.Interface, config string, useCrdConfiguration bool) *LB {
```
ctreatma, line 65: If it's important _not_ to change this function signature, and if the `metallb:///<configNamespace>/<configMapName>` format for the `config` string was invented by us, we could alternatively introduce a different config string format for newer MetalLB.

metal/loadbalancers.go
```diff
@@ -49,7 +49,7 @@ type loadBalancers struct {
 	eipTag                string
 }
 
-func newLoadBalancers(client *packngo.Client, k8sclient kubernetes.Interface, projectID, metro, facility, config string, localASN int, bgpPass, annotationNetwork, annotationLocalASN, annotationPeerASN, annotationPeerIP, annotationSrcIP, annotationBgpPass, eipMetroAnnotation, eipFacilityAnnotation, nodeSelector, eipTag string) (*loadBalancers, error) {
+func newLoadBalancers(client *packngo.Client, k8sclient kubernetes.Interface, projectID, metro, facility, config string, localASN int, bgpPass, annotationNetwork, annotationLocalASN, annotationPeerASN, annotationPeerIP, annotationSrcIP, annotationBgpPass, eipMetroAnnotation, eipFacilityAnnotation, nodeSelector, eipTag string, useCRDForMetalLB bool) (*loadBalancers, error) {
```
ctreatma, line 52: I don't love how this one flag for a specific upstream change in one load balancer implementation bubbles all the way up through each layer, but maybe that's a weird thing to focus on.

metal/config_test.go
```diff
@@ -105,6 +122,32 @@ func Test_getMetalConfig(t *testing.T) {
 			want:    Config{},
 			wantErr: true,
 		},
+		{
```
ctreatma, line 125: I updated unit tests but haven't done an e2e test yet in a real cluster.

Lirt:
Hi, please see - https://github.com/equinix/cloud-provider-equinix-metal/issues/385#issuecomment-1457830600.

The configmap/namespace parser is not working correctly. In case there is only namespace defined (`METAL_LOAD_BALANCER: "metallb:///kube-system"` as per docs) the `namespace` variable will be set to default value (`metallb-system`). 

This needs to be fixed as well.

cprivitere:
So metallb is throwing this warning:
```
metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
```

cprivitere:
> So metallb is throwing this warning:
> 
> ```
> metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
> ```

This may be a false positive, I've checked and it seems we're using IPAddressPool.

cprivitere (approved):

PR #410, opened 2023-05-09
fix: correctly use the eip facility annotation

### LINKED ISSUE #409: EIP Facility Annotation ignored
According to the docs, there is a (configurable) annotation for Service resources to select the facility from which an EIP can be associated, which by default is `metal.equinix.com/eip-facility`.

From my testing, this annotation is ignored, while the metro-based `metal.equinix.com/eip-metro` annotation works as expected.

TimJones:
Fix the load balancer ignoring the EIP facilicy selection annotation by actually passing the correct annotation configuration to the new load balancer function.

Fixes #409


cprivitere (approved):
lgtm

PR #417, opened 2023-05-31
doc: Update docs to be inline with templates

cprivitere:
This is a first attempt to get our docs in line with the [Kubernetes Templates](https://github.com/kubernetes/kubernetes-template-project/blob/main/README.md) as part of our effort to get the project migrated to kubernetes-sigs.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,192 @@
-   Copyright 2018-2021 Equinix, Inc.
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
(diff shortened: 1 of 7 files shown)


RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+This is normally performed by our CI system. However, there are important steps to take first.
+
+## How to Cut a Release
+
+In order to cut a release, you must:
+
+1. Commit the changes.
+1. Push out your branch, open a PR and merge the changes
+1. Wait for the Continuous Integration github action to finish running
+1. Tag the release with `git tag -a vX.Y.z -m "Message"
```
displague, line 16:
```suggestion
1. Tag the release with `git tag -a vX.Y.z -m "Message"`
```

RELEASE.md
```diff
@@ -0,0 +1,24 @@
...
+1. Push out the tag
+
+## How A Release Happens
+
+* GitHub Actions detects a new tag has been pushed
+* CI builds docker images for each supported architecture as well as a multi-arch manifest, and tags it with the semver tag of the release, e.g. `v0.4.0`
+* CI creates the release in `out/release`, the equivalent of `make release`
```
displague, line 23: would we be changing our Makefile to abide by this? (no `release` task defined today)

displague (dismissed):
lgtm - nits aside

LICENSE
```diff
@@ -1,4 +1,192 @@
-   Copyright 2018-2021 Equinix, Inc.
+                                 Apache License
```
ctreatma, line 1: Do we need anyone to sign off on this license change?

ctreatma: Nevermind, it's the same license, your PR just adds the rest of the license text.

displague: Moving from Apache to Apache. The difference is the copyright range and name, which in this PR is updated with the LICENSE template text. We should add Kubernetes Authors too (unless there is a newer convention).

Here's an example: https://github.com/linode/linode-cloud-controller-manager/blob/master/LICENSE#L189-L192

displague: It's also pretty common, to leave this as the template text.. https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/LICENSE#L189

displague (approved):

PR #479, opened 2023-11-03
fix: hook the EMLB client up to our debug env var

ctreatma:
This updates the Equinix Metal Load Balancer integration to enable HTTP client debug logs when the `PACKNGO_DEBUG` environment variable is set.  In the future we may introduce other ways of enabling debug logs for the provider, but this change enables users to debug the EMLB parts of the provider using the same approach they use to debug the rest of the provider.


displague (approved):

displague:
Should this also include auth key redaction? (this will look different since the `Authorization` header is what will need redacting)

Issue #439, opened 2023-08-04
Replace packngo with metal-go


displague:
https://github.com/equinix-labs/metal-go is an OAS3 generated SDK that is the successor to the hand curated https://github.com/packethost/packngo SDK.

The latter is being deprecated once key integrations including this project, https://github.com/kubernetes-sigs/cluster-api-provider-packet/pull/527, metal-cli, and terraform-provider-equinix have successfully adopted the new SDK (thus proving its completeness and accuracy).  

Related to https://github.com/equinix-labs/metal-go/issues/99

ctreatma:
Just noting here, `packngo` was instead replaced with [`equinix-sdk-go`](https://github.com/equinix/equinix-sdk-go) because that SDK is roughly identical to `metal-go` and is maintained (`metal-go` was experimental and is now deprecated itself).

PR #486, opened 2023-12-13
chore: removed packngo and used metal-go apis

aayushrangwala:
fixes: #439


ctreatma:
There should be a `go.mod` change here to  pull in the new SDK instead of the old one.

Now that [equinix-sdk-go[(https://github.com/equinix/equinix-sdk-go) exists, we should adopt that instead of metal-go.

metal/devices.go
```diff
@@ -17,13 +15,13 @@ import (
 )
 
 type instances struct {
-	client  *packngo.Client
+	client  *metal.DevicesApiService
```
ctreatma, line 18: There's a devices_test.go file as well that will need to be updated to use equinix-sdk-go instead of packngo

aayushrangwala: Yes updated

ctreatma:
@aayushrangwala The [PR test workflow](https://github.com/kubernetes-sigs/cloud-provider-equinix-metal/actions/runs/7340676922/job/19987075237?pr=486) is failing with the error `failed to enable BGP on project cd5ad69d-5460-46f8-8268-963262f2b8c2: undefined response type`.  Are you able to reproduce that failure locally?

aayushrangwala:
> @aayushrangwala The [PR test workflow](https://github.com/kubernetes-sigs/cloud-provider-equinix-metal/actions/runs/7340676922/job/19987075237?pr=486) is failing with the error `failed to enable BGP on project cd5ad69d-5460-46f8-8268-963262f2b8c2: undefined response type`. Are you able to reproduce that failure locally?

@ctreatma Yes I am fixing the tests for the same. Previously there was packngo packetServer which was structured for tests/interfaces and accordingly the tests were written. Changing the structure to implement bare minimum test server and fixing tests

aayushrangwala:
@ctreatma @displague Fixed all the tests. Please review and approve if looks good

metal/bgp.go
```diff
@@ -47,35 +51,42 @@ func (b *bgp) enableBGP() error {
...
 	// we did not have a valid one, so create it
-	req := packngo.CreateBGPConfigRequest{
-		Asn:            b.localASN,
-		Md5:            b.bgpPass,
+	req := metal.BgpConfigRequestInput{
+		Asn:            int32(b.localASN),
+		Md5:            &b.bgpPass,
 		DeploymentType: "local",
-		UseCase:        "kubernetes-load-balancer",
+		UseCase:        pointer.String("kubernetes-load-balancer"),
```
ctreatma, line 65: the `metalv1` package has helpers for this so you can avoid pulling in an extra dependency; since you've aliased the package as `metal` here, the call would be [`metal.PtrString("kubernetes-load-balancer")`](https://github.com/equinix/equinix-sdk-go/blob/main/services/metalv1/utils.go#L39)

metal/cloud.go
```diff
@@ -4,11 +4,10 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/packethost/packngo"
-
+	metal "github.com/equinix/equinix-sdk-go/services/metalv1"
 	cloudprovider "k8s.io/cloud-provider"
 	"k8s.io/component-base/version"
-	"k8s.io/klog/v2"
```
ctreatma, line 11: This probably shouldn't change, but I don't know if there was an error or something that forced this change?

metal/cloud_test.go
```diff
@@ -211,14 +209,89 @@ func TestHasClusterID(t *testing.T) {
 }
 
 // builds an Equinix Metal client
-func constructClient(authToken string, baseURL *string) *packngo.Client {
-	client := retryablehttp.NewClient()
+func constructClient(authToken string, baseUrl *string) *metal.APIClient {
+	configuration := &metal.Configuration{
```
ctreatma, line 213: Instead of setting all of these options here, you could do `configuration := metal.NewConfiguration()`, and then if `baseURL != nil` you can set `configuration.Servers` to point to `*baseURL`.

metal/cloud_test.go
```diff
@@ -211,14 +209,89 @@ func TestHasClusterID(t *testing.T) {
...
 	}
-	return packngo.NewClientWithAuth(ConsumerToken, authToken, client.StandardClient())
+
+	configuration.Servers = servers
+	configuration.AddDefaultHeader("X-Auth-Token", authToken)
+	configuration.UserAgent = fmt.Sprintf("cloud-provider-equinix-metal/%s %s", version.Get(), configuration.UserAgent)
+	return metal.NewAPIClient(configuration)
+}
+
+func (s *MockMetalServer) CreateHandler() http.Handler {
```
ctreatma, line 242: This is good stuff!  It might be nice to move the `MockMetalServer` definition and this `CreateHandler` function to a separate file so it's easier to generalize & extract in the future; I could imagine the `MockMetalServer` taking a map of route => HTTP method => handler function, but we might need to figure out scope binding so that the handlers functions could reference the mock server's data store.

ctreatma (commented):
The updated tests look great.  I see a couple changes that look like they might not be necessary for the migration, but not sure about that.  I don't have write access to this repo in any case, so we ultimately need @cprivitere to be included in the review process as well.

go.mod
```diff
@@ -113,8 +116,12 @@ require (
...
 	k8s.io/kms v0.26.10 // indirect
 	k8s.io/kube-openapi v0.0.0-20221012153701-172d655c2280 // indirect
-	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448 // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.0.37 // indirect
 	sigs.k8s.io/json v0.0.0-20220713155537-f223a00ba0e2 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 )
+
+replace (
+	k8s.io/kubectl => k8s.io/kubectl v0.28.4
```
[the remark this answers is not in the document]
aayushrangwala: `k8s.io/kubectl` It was pointing to `v0.0.0` and hence failing to compile

metal/bgp.go
aayushrangwala: Ah right. Didnt know. Changed

metal/cloud.go
aayushrangwala: It got changed via go mod tidy I guess. Reverted

metal/cloud_test.go
aayushrangwala: Yes, a store interface can be created which will be internal to testing and inmemory datastore can be implentend for the interface. This will bind which type of store is used for the server.

go.mod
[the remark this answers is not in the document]
cprivitere: These don't feel right to me. I've not needed them before or encountered them elsewhere.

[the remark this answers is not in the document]
aayushrangwala: Yes removed and tried to run binary locally. Its building and also the tests, linters are passing. So maybe that was a temp issue on my machine

cprivitere (commented):
/approve

cprivitere (commented):
/lgtm

PR #581, opened 2024-11-15
Remove deitch as owner

deitch:
I haven't really maintained this in a very long time. Removing myself from the list.


displague:
Thanks for all your work on this and related projects!

PR #613, opened 2025-03-21
Drop Equinix Metal Load Balancer

tylerauerbeck:
As the Equinix Metal Load Balancer service has been sunset -- this removes the EMLB configurations that were put in place to support this feature from CPEM.


displague:
"lgtm" was to get PROW unblocked.

I have concerns about this PR, and I'm not sure I can give it the review it deserves.

This will be hard to test since the service has been sunset. 

Even if we ignore that, we should consider side effects on users who have previously deployed a load balancer through CPEM (even if the underlying instance is gone). If this is impractical, we can smooth over gaps with documentation in the release notes and README.md: "if you had X, here's how to unwind it before you update to a post-lbaas CPEM". 

The best case would be to offer a migration guide, but I think reviewer constraints prevent that.

One thing to consider is whether Services or other resources managed through this CPEM LBaaS code were assigned finalizers or other configuration options that will need to be manually removed in the absence of a supporting, dedicated, controller code.

PR #650, opened 2025-08-28
update kubernetes dependencies to 1.34

jimohabdol:
- Update GOVER to 1.24.0 and update any related dependencies that are needed by ks8 1.34


elmiko:
i don't think we need to expand the test coverage here, especially since we are deprecating this repository.

i will give a full review tomorrow, but i am concerned to see a test for "atoi", we do not need that.

jimohabdol:
> i don't think we need to expand the test coverage here, especially since we are deprecating this repository.
> 
> i will give a full review tomorrow, but i am concerned to see a test for "atoi", we do not need that.

I will remove those tests since we don't need it

